Issue #25, opened 2017-04-21
Rewrite TransferFromDeviceDialog as Gtk.Dialog


elementaryBot:
Dialogs should be a Gtk.Dialog subclass, not Gtk.Window

Launchpad Details: [#LP1310481](https://bugs.launchpad.net/bugs/1310481) Daniel Fore - 2014-04-21 07:04:19 +0000

Bat41:
Is this dialog still useful? I'm looking at the code, and the only place it seems to be used is in `Noise.DeviceView.showImportDialog`, which is never called.

Bat41:
Closed by #120

PR #39, opened 2017-06-05
Use Gtk.Dialog for dialogs

Bat41:
Rewrote `SyncWarningDialog` as a subclass of `Gtk.Dialog`.

There is also the `TranferFromDeviceDialog`, but as I asked in #25 I don't know if we even need to keep this part of the code…

### DIFF
--- a/po/nl.po
+++ b/po/nl.po
@@ -8,15 +8,16 @@ msgstr ""
 "Project-Id-Version: beat-box\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2016-12-18 14:43+0100\n"
-"PO-Revision-Date: 2017-07-01 11:14+0000\n"
-"Last-Translator: Noah Loomans <noahloomans@gmail.com>\n"
-"Language-Team: Dutch <https://l10n.elementary.io/projects/music/music/nl/>\n"
+"PO-Revision-Date: 2017-08-05 16:14+0000\n"
+"Last-Translator: Dieter Debast <dieter_debast@hotmail.com>\n"
+"Language-Team: Dutch <https://weblate.elementary.io/projects/music/music/nl/>"
+"\n"
 "Language: nl\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=n != 1;\n"
-"X-Generator: Weblate 2.14\n"
+"X-Generator: Weblate 2.15\n"
 "X-Launchpad-Export-Date: 2017-03-07 05:52+0000\n"
 
 #: ../core/LibrariesManager.vala:143
@@ -379,7 +380,7 @@ msgstr "Voeg intelligente afspeellijst toe"
 #. / Do not remove '%s'. It's a placeholder for selected equalizer preset name.
 #: ../src/Widgets/StatusBar.vala:189
 msgid "Equalizer: %s"
-msgstr ""
+msgstr "Equalizer: %s"
 
 #: ../src/Widgets/StatusBar.vala:199
 msgid "Hide"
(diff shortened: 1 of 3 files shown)


Bat41:
I just noticed that the commit history is not very clean… I'll try to change that.

src/Dialogs/SyncWarningDialog.vala
```diff
@@ -73,75 +79,58 @@ public class Noise.SyncWarningDialog : Gtk.Window {
...
-        string title_text = "";
-        if (to_remove.size > 1) {
-            title_text = _("Sync will remove %i items from %s").printf (to_remove.size, d.getDisplayName ());
-        }
-        else {
-            title_text = _("Sync will remove 1 item from %s").printf (d.getDisplayName ());
-        }
+        string title_text = to_remove.size > 1
+            ? _("Sync will remove %i items from %s").printf (to_remove.size, d.getDisplayName ())
+            : _("Sync will remove 1 item from %s").printf (d.getDisplayName ());
```
jeremypw: Can you use ngettext or a variant thereof to handle the plural forms here?  What if to_remove.size = 0 (not sure if this is possible in practice)?

Bat41: I think the message will just be `Sync will remove 0 item from [Device Name Here]`.

jeremypw: ... which is not grammatical.  Zero should take the plural form (in English).  I think ngettext takes care of this.

Bat41: Oh, yes, I am not a native English speaker, Id didn't know this rule. But, yes, I'll just use ngettext.

src/Dialogs/SyncWarningDialog.vala
```diff
@@ -26,44 +26,54 @@
...
 
-public class Noise.SyncWarningDialog : Gtk.Window {
+public class Noise.SyncWarningDialog : Gtk.Dialog {
+    private enum ResponseId {
+        IMPORT_MEDIA = 1,
+        CONTINUE,
+        STOP
+    }
+
     Device d;
```
jeremypw: Make this a construction property and rename appropriately.  e.g. public Device device {get; construct;}

src/Dialogs/SyncWarningDialog.vala
```diff
@@ -26,44 +26,54 @@
...
+        continue_button = add_button (_("Continue Syncing"), ResponseId.CONTINUE);
+        stop_button = add_button (_("Stop Syncing"), ResponseId.STOP);
+        response.connect (on_response);
+
+        import_media_button.sensitive = !libraries_manager.local_library.doing_file_operations ();
+        continue_button.sensitive = !libraries_manager.local_library.doing_file_operations ();
+    }
+
+    public SyncWarningDialog (Device d, Gee.Collection<Media> to_sync, Gee.Collection<Media> removed) {
         this.d = d;
```
jeremypw: Replace with Object (device: d);

src/Dialogs/SyncWarningDialog.vala
```diff
@@ -26,44 +26,54 @@
...
-        this.set_transient_for(App.main_window);
-        this.destroy_with_parent = true;
-        
-        set_default_size(475, -1);
-        resizable = false;
-        
-        content = new Gtk.Box(Gtk.Orientation.VERTICAL, 10);
-        padding = new Gtk.Box(Gtk.Orientation.HORIZONTAL, 20);
-        
+        Gtk.Box content = get_content_area ();
```
jeremypw: This and most of the following code in the constructor can be transferred to the construct block.   Only retain in the constructor code that requires reference, direct or indirect, to parameters "to_sync" and "remove".

Bat41:
I moved everything to the `construct` block (and transformed `to_sync` and `to_remove` to properties), it was the simplest solution.

I also think we will need to update translation files to make `ngettext` work correctly, but I don't know how to achieve that.

jeremypw:
Conflicts with master need to be resolved before further review.

jeremypw:
As it seems I do not have a device that Noise recognises as as suitable type to synchronise, I cannot review the function of the new code, although it compiles cleanly.

You should consider transferring the code that deals with the dialog responses to the owner of the dialog, so that device.synchronise is only called there.

matzipan:
There's an outstanding conflict that must be resolved, @Bat41

Issue #54, opened 2017-08-10
Can't shuffle tracks that aren't in the library


Nine-H:
I opened a bunch of tracks I scraped off soundcloud, and they were added to the queue, which is great, but the shuffle button has no effect on the queue.

I think the most elegant solution is putting a context option on queued tracks to shuffle them up: this would make it easy to add a couple of albums and blend them into a temp playlist while not giving an unexpected behavior to people that want to cue up a track to play after the currently playing track.

Issue #42, opened 2017-06-30
Shuffle is not confined to playlist / column view selection


bbuhler:
**Current behavior**
```gherkin
Given I have created a playlist
And the shuffle mode is enabled
When I play a song from this playlist
And click on next button
Then any other song of my library plays
```

**Expected behavior**
```gherkin
Given I have created a playlist
And the shuffle mode is enabled
When I play a song from this playlist
And click on next button
Then one of the songs of the playlist should be played
But not any other song of my library
```

This is also the case when I select an album / genre / artist in column view.

The bug was already reported and confirmed on [launchpad](https://bugs.launchpad.net/noise/+bug/1507047).
It is still present in the latest version (0.4.0.3).

PR #65, opened 2017-09-03
Change the queue mechanism to make it less confusing

### LINKED ISSUE #2: The queue mechanism is a bit confusing
The way the queue works is a bit confusing for several reasons.

1)  it is populated manually (right click enqueue). 
Yet when the user double clicks on a media, the media is played automatically but the queue is not updated: the media coming next to the double clicked media will play (but will not show in the queue)

2) how should play/next/previous buttons work?
Should they play the currently selected track or play the queue playlist

Also, the queue is not persisted: it is lost when the application exits.

The iTunes queue is nice insofar it always shows what will the next track be and let the user re order and remove tracks.

Note: the technical solution must scale for large libraries.

Launchpad Details: [#LP1382879](https://bugs.launchpad.net/bugs/1382879) PerfectCarl - 2014-10-18 23:31:33 +0000

Bat41:
Fix #2, #54 and #42 

Now, when you double-click on a song, it is played and other songs of
the current list are added to the queue.

Also improved the style of some parts of the code, and changed many
Gee.HashMap<int, Media> to ArrayList<int, Media> (they were acting as
arrays, I don't know why they were maps).


core/Playlists/Playlist.vala
```diff
@@ -25,18 +25,18 @@
  *
  * Authored by: Corentin Noël <corentin@elementary.io>
  */
- 
+
 public abstract class Noise.Playlist : Object {
     public signal void request_play ();
-    public virtual Gee.ArrayQueue<Media> medias { get; internal set; }
+    public virtual Gee.ArrayList<Media> medias { get; internal set; }
```
matzipan: I think here you can change it like: 
```
get;
internal set;
default = new Gee.ArrayList<Media> ();
``` 

This way we avoid having the code in the constructor. You could use a similar pattern to all the code in the constructor.

matzipan: Also `media` is the plural word for `medium`. So there is no such thing as `medias`, perhaps the variable should be set to `media` and the class name to `Medium`?

core/Settings.vala
```diff
@@ -116,7 +116,7 @@ namespace Noise.Settings {
             base ("org.pantheon.noise.equalizer");
         }
 
-        public Gee.Collection<Noise.EqualizerPreset> getPresets () {
+        public Gee.Collection<Noise.EqualizerPreset> get_presets () {
```
matzipan: thanks for this

plugins/Devices/CDRom/CDView.vala
```diff
@@ -28,115 +28,115 @@
...
+
         show_all ();
     }
-    
+
     public void cd_initialised () {
         cd_playlist.add_medias (dev.get_medias ());
         if ( cd_playlist.is_empty () == false) {
-            var m = cd_playlist.medias.peek ();
+            var m = cd_playlist.medias[0];
```
matzipan: You have introduced the `get` and `set` methods in the `Noise.Playlist` class, yet you are referencing `medias` directly, shouldn't `medias` be turned into a private attribute?

src/Views/ListView/ListView.vala
```diff
@@ -269,15 +269,11 @@ public class Noise.ListView : ContentView, Gtk.Box {
     }
 
     public Gee.Collection<Media> get_media () {
-        var media_list = new Gee.ArrayQueue<Media> ();
-        media_list.add_all (list_view.get_table ().values);
-        return media_list;
+        return list_view.get_table ();
```
matzipan: The previous code was a soft copy of the list returned by `get_table()`. This is usually done as a defensive mechanism when possibility that the list will be mutated by the outside code, to prevent the outside code from modifying the  internal list of the `list_view`. Are you sure this will not introduce side effects and issues when the caller of `get_media` removes or adds things to the list?

src/Views/ListView/ListView.vala
```diff
@@ -269,15 +269,11 @@ public class Noise.ListView : ContentView, Gtk.Box {
...
     }
 
     public Gee.Collection<Media> get_visible_media () {
-        var media_list = new Gee.ArrayQueue<Media> ();
-        media_list.add_all (list_view.get_visible_table ().values);
-        return media_list;
+        return list_view.get_visible_table ();
```
matzipan: Same here as above.

src/Views/ListView/ListView.vala
```diff
@@ -385,41 +381,32 @@ public class Noise.ListView : ContentView, Gtk.Box {
         return status_text;
     }
 
-    private void view_search_func (string search, Gee.HashMap<int, Media> table, Gee.HashMap<int, Media> showing) {
+    private void view_search_func (string search, Gee.ArrayList<Media> table, Gee.ArrayList<Media> showing) {
```
matzipan: I'm not familiar with the code, but do you think there was any reason why `HashMaps` were chosen previously?

src/Views/ListView/Lists/GenericList.vala
```diff
@@ -104,7 +105,7 @@ public abstract class Noise.GenericList : FastView {
 
         parent_wrapper.library.media_updated.connect (media_updated);
 
-        App.player.current_cleared.connect (current_cleared);
+        App.player.queue_cleared.connect (current_cleared);
```
matzipan: What's the difference between the two and how does this relate with the issues being fixed?

plugins/Devices/CDRom/CDView.vala
Bat41: Yes, I think I'll have to introduce a few more methods to `Playlist`s (at least one to get `media.size`), but it would be better.

src/Views/ListView/ListView.vala
Bat41: Oh, I didn't thought about that.

src/Views/ListView/ListView.vala
Bat41: No, I searched, but it looks like they were used like arrays. That's strange…

core/EqualizerPreset.vala
```diff
@@ -29,52 +29,55 @@
...
 
 public class Noise.EqualizerPreset : Object {
-
     public string name = "";
     public Gee.ArrayList<int> gains = new Gee.ArrayList<int> ();
 
     public bool is_default { get; set; default = false; }
 
     public EqualizerPreset.basic (string name) {
         this.name = name;
```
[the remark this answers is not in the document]
matzipan: In the current file and all subsequent files.

src/Views/ListView/Lists/GenericList.vala
Bat41: Before my changes, you had two different list in `PlaybackManager` : `current` and `queue`. What I did is that I merged these two list into `queue`, that's why I renamed this signal.

matzipan:
Actually make sure you also change the class construction method to GObject construction, the method Jeremy has talked to you in your old Dialog PR.

core/Playlists/Playlist.vala
Bat41: It's a big change, I would prefer to do it in a dedicated PR.

matzipan:
When running the code I'm getting this: 
```
[WARNING 23:42:19.795664] [GLib-GObject] invalid (NULL) pointer instance
[FATAL 23:42:19.795724] [GLib-GObject] g_signal_connect_object: assertion 'G_TYPE_CHECK_INSTANCE (instance)' failed
[FATAL 23:42:19.795770] noise_list_view_construct: assertion 'tvs != NULL' failed
[FATAL 23:42:19.795806] [GLib-GObject] g_object_ref_sink: assertion 'G_IS_OBJECT (object)' failed
**
ERROR:/home/matzipan/Workspace/music/src/Views/Wrappers/PlaylistViewWrapper.vala:76:noise_playlist_view_wrapper_constructor: code should not be reached

```

Full backtrace:
```
#0  0x00007ffff43b8428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
#1  0x00007ffff43ba02a in __GI_abort () at abort.c:89
#2  0x00007ffff4d43d65 in g_assertion_message () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#3  0x00007ffff4d43dfa in g_assertion_message_expr () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#4  0x00000000004dc35c in noise_playlist_view_wrapper_constructor ()
#5  0x00007ffff4ffaad8 in ?? () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#6  0x00007ffff4ffc1b5 in g_object_new_valist () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#7  0x00007ffff4ffc521 in g_object_new () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#8  0x00000000004dac10 in noise_playlist_view_wrapper_construct ()
#9  0x00000000004dac57 in noise_playlist_view_wrapper_new ()
#10 0x000000000043eea3 in noise_library_window_create_smartplaylist_source_list ()
#11 0x000000000043dcaa in noise_library_window_add_smartplaylist ()
#12 0x0000000000438fb1 in noise_library_window_load_playlists ()
#13 0x000000000043703a in noise_library_window_build_ui ()
#14 0x00000000004304e4 in noise_app_real_activate ()
#15 0x00007ffff4ff51d4 in ?? () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#16 0x00007ffff500f9a6 in g_signal_emit_valist () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#17 0x00007ffff501008f in g_signal_emit () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#18 0x00007ffff6e0bae6 in ?? () from /usr/lib/x86_64-linux-gnu/libgio-2.0.so.0
#19 0x00007ffff6e0bcd2 in g_application_run () from /usr/lib/x86_64-linux-gnu/libgio-2.0.so.0
#20 0x00007ffff73cf450 in granite_application_run () from /usr/lib/x86_64-linux-gnu/libgranite.so.3
#21 0x000000000042fede in _vala_main ()
#22 0x000000000042ff79 in main ()
```

matzipan:
Some functional testing results: 

Initial state: Queue empty
I have a playlist of 4 items:
screenshot from 2017-09-09 21 07 00

When I click on an item, the queue becomes 2*(number of items), in this case 8:
screenshot from 2017-09-09 21 07 15

And there are a few more issues, the currently playing song is not the first one in the list and if I click on any of the songs in the queue, the list becomes empty.

src/Widgets/FastView/FastList.vala
```diff
@@ -64,6 +64,11 @@ public class Noise.FastView : Gtk.TreeView {
     }
 
     construct {
+        columns = new Gee.ArrayList<Type> ();
+        foreach (var type in ListColumn.get_all ()) {
```
matzipan: Any reason why you moved this from GenericList?

src/Widgets/FastView/FastList.vala
```diff
@@ -64,6 +64,11 @@ public class Noise.FastView : Gtk.TreeView {
     }
 
     construct {
+        columns = new Gee.ArrayList<Type> ();
```
matzipan: This should be in the default field.

src/Widgets/FastView/FastList.vala
Bat41: Yes, before, this constructor got called before creating `GenericList`s, but the columns were empty. If i left it in the other constructor, I would had to duplicate some code from here (like creating the `FastModel`). (I hope my explanation was clear).

Bat41:
I can reproduce the three problems you had, I'll try to fix them soon.

src/Widgets/FastView/FastList.vala
matzipan: It's not entirely clear, to be honest.

Bat41: Okay, I'll try to tell it differently. Before, when you were creating a `GenericList` instance, it started by calling the parent class (`FastView`) construct block, before calling the GenericList's construct block itself.

The FastView's construct block was using `columns`, because when we were using C#-style construction, we could just pass them from GenericList with `base` after having initialized them. But after changing code to use GObject construction style, it was `null`, and it caused a lot of problems.

I hope it was clearer this time (I'm sorry, my English isn't perfect, I can't say everything I would like to).

matzipan: Your explanation was right the first time around. I did not realize `GenericList` is a subclass of `FastView`. I thought it was a superclass.

matzipan:
Hey @Bat41  any updates?

src/Widgets/FastView/FastList.vala
Bat41: Well, I tried, but it looks like it just breaks everything…

Bat41:
Hey @matzipan, I think I have fixed all the issues you found. And about that:

> If I change to the second song, shouldn't it become the first song? And shouldn't there be 3 songs after it? If I change to the last song, shouldn't it become the first and 3 random songs after?

In my opinion, it could be confusing to see the list change after each song. For example if you want to go back to a song you listened to a few minutes ago, and you have to go to the end of the list while it has been played before the current one. Not sure if I'm clear. :grin:

PR #66, opened 2017-09-05
URI Filter : Search and SmartPlaylist

mohansharma-me:
Now users can search through URI. So they can create playlist for particular directory.
For example if I want to play music from particular directory then it can be done by searching "uri:/path/to/directory" or if I want to create a playlist from directory /path/to/directory then I will follow this steps:
1. Search "uri:/path/to/directory"
2. Select all results
3. Click : Right-click Menu > Add to playlist
4. Name it.

### DIFF
--- a/src/Dialogs/SmartPlaylistEditor.vala
+++ b/src/Dialogs/SmartPlaylistEditor.vala
@@ -293,7 +293,7 @@ public class Noise.SmartPlaylistEditorQuery : GLib.Object {
 
         if (needs_value (q.field)) {
             if(q.field == SmartQuery.FieldType.URI) {
-                value_entry.text = q.value.get_string ().replace("%20", " ");
+                value_entry.text = Uri.unescape_string(q.value.get_string ());
             } else {
                 value_entry.text = q.value.get_string ();
             }
@@ -330,7 +330,7 @@ public class Noise.SmartPlaylistEditorQuery : GLib.Object {
         if (needs_value ((SmartQuery.FieldType)field_combobox.get_active ())) {
             var value = Value (typeof (string));
             if(rv.field == SmartQuery.FieldType.URI) {
-                value.set_string (value_entry.text.replace(" ", "%20"));
+                value.set_string (Uri.escape_string(value_entry.text, "/"));
             } else {
                 value.set_string (value_entry.text);
             }
--- a/src/Dialogs/SmartPlaylistEditor.vala
+++ b/src/Dialogs/SmartPlaylistEditor.vala
@@ -292,8 +292,8 @@ public class Noise.SmartPlaylistEditorQuery : GLib.Object {
         comparator_combobox.set_active ((int)q.comparator);
 
         if (needs_value (q.field)) {
-            if(q.field == SmartQuery.FieldType.URI) {
-                value_entry.text = Uri.unescape_string(q.value.get_string ());
+            if (q.field == SmartQuery.FieldType.URI) {
+                value_entry.text = Uri.unescape_string (q.value.get_string ());
             } else {
                 value_entry.text = q.value.get_string ();
             }
@@ -329,8 +329,8 @@ public class Noise.SmartPlaylistEditorQuery : GLib.Object {
         rv.comparator = comparators.get (comparator_combobox.get_active ());
         if (needs_value ((SmartQuery.FieldType)field_combobox.get_active ())) {
             var value = Value (typeof (string));
-            if(rv.field == SmartQuery.FieldType.URI) {
-                value.set_string (Uri.escape_string(value_entry.text, "/"));
+            if (rv.field == SmartQuery.FieldType.URI) {
+                value.set_string (Uri.escape_string (value_entry.text, "/"));
             } else {
                 value.set_string (value_entry.text);
             }
--- a/src/Dialogs/SmartPlaylistEditor.vala
+++ b/src/Dialogs/SmartPlaylistEditor.vala
@@ -285,13 +285,18 @@ public class Noise.SmartPlaylistEditorQuery : GLib.Object {
         field_combobox.append_text (_("Skipcount"));
         field_combobox.append_text (_("Title"));
         field_combobox.append_text (_("Year"));
+        field_combobox.append_text (_("URI"));
 
         field_combobox.set_active ((int)q.field);
         debug ("setting filed to %d\n", q.field);
         comparator_combobox.set_active ((int)q.comparator);
 
         if (needs_value (q.field)) {
-            value_entry.text = q.value.get_string ();
+            if(q.field == SmartQuery.FieldType.URI) {
+                value_entry.text = q.value.get_string ().replace("%20", " ");
+            } else {
+                value_entry.text = q.value.get_string ();
+            }
         } else if (q.field == SmartQuery.FieldType.RATING) {
             _valueRating.rating = q.value.get_int ();
         } else {
@@ -324,7 +329,11 @@ public class Noise.SmartPlaylistEditorQuery : GLib.Object {
         rv.comparator = comparators.get (comparator_combobox.get_active ());
         if (needs_value ((SmartQuery.FieldType)field_combobox.get_active ())) {
             var value = Value (typeof (string));
-            value.set_string (value_entry.text);
+            if(rv.field == SmartQuery.FieldType.URI) {
+                value.set_string (value_entry.text.replace(" ", "%20"));
+            } else {
+                value.set_string (value_entry.text);
+            }
             rv.value = value;
         } else if (field_combobox.get_active () == SmartQuery.FieldType.RATING) {
             var value = Value (typeof (int));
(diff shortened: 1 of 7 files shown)


src/LocalBackend/LocalLibrary.vala
```diff
@@ -565,11 +565,33 @@ public class Noise.LocalLibrary : Library {
...
+                    
+                    /* Old SQL query generation code */
+                    fields = {"title", "artist", "composer", "album_artist", "album", "grouping", "comment"};
+        
+                    foreach (var field in fields) {
+                        var id_field = sql.add_id (field);
+                        var id_value = sql.add_expr_value (null, "%"+search+"%");
+                        ids += sql.add_cond (Gda.SqlOperatorType.LIKE, id_field, id_value, 0);
+                    }
+
```
artemanufrij: you don't need to use foreach statement twice. Just one time after else block.

mohansharma-me:
I pushed required changes. :+1:

mohansharma-me:
I've also added URI filter option in SmartPlaylist Editor so users can also create smartplaylist for directory.

src/Dialogs/SmartPlaylistEditor.vala
```diff
@@ -285,13 +285,18 @@ public class Noise.SmartPlaylistEditorQuery : GLib.Object {
...
+        field_combobox.append_text (_("URI"));
 
         field_combobox.set_active ((int)q.field);
         debug ("setting filed to %d\n", q.field);
         comparator_combobox.set_active ((int)q.comparator);
 
         if (needs_value (q.field)) {
-            value_entry.text = q.value.get_string ();
+            if(q.field == SmartQuery.FieldType.URI) {
+                value_entry.text = q.value.get_string ().replace("%20", " ");
```
matzipan: Is there a reason why [GLib.Uri](https://valadoc.org/glib-2.0/GLib.Uri.html) cannot be used in this case?

src/Dialogs/SmartPlaylistEditor.vala
```diff
@@ -324,7 +329,11 @@ public class Noise.SmartPlaylistEditorQuery : GLib.Object {
         rv.comparator = comparators.get (comparator_combobox.get_active ());
         if (needs_value ((SmartQuery.FieldType)field_combobox.get_active ())) {
             var value = Value (typeof (string));
-            value.set_string (value_entry.text);
+            if(rv.field == SmartQuery.FieldType.URI) {
+                value.set_string (value_entry.text.replace(" ", "%20"));
```
matzipan: Same here: Is there a reason why [GLib.Uri](https://valadoc.org/glib-2.0/GLib.Uri.html) cannot be used in this case?

src/LocalBackend/LocalLibrary.vala
```diff
@@ -565,10 +565,20 @@ public class Noise.LocalLibrary : Library {
                 sql.select_add_field ("rowid", null, null);
                 Gda.SqlBuilderId[] ids = null;
 
-                string[] fields = {"title", "artist", "composer", "album_artist", "album", "grouping", "comment"};
+                /* Enabling directory/path/uri search through SearchEntry */
+                string new_search = search;
+                string[] fields = {};
+                if(search.has_prefix("uri:")) {
+                    new_search = search.substring(4);
+                    new_search = new_search.replace(" ", "%20");
```
matzipan: Same here: Is there a reason why [GLib.Uri](https://valadoc.org/glib-2.0/GLib.Uri.html) cannot be used in this case?

src/LocalBackend/LocalLibrary.vala
```diff
@@ -565,10 +565,20 @@ public class Noise.LocalLibrary : Library {
                 sql.select_add_field ("rowid", null, null);
                 Gda.SqlBuilderId[] ids = null;
 
-                string[] fields = {"title", "artist", "composer", "album_artist", "album", "grouping", "comment"};
+                /* Enabling directory/path/uri search through SearchEntry */
```
matzipan: Can you clarify what the role of this code is and why uri can't be included with the other fields? Can you have a search by uri and by title at the same time?

mohansharma-me:
Actually reason behind using raw string-replace function to encode/decode the path is that in media library path "/home/someone/music directory" is stored as "/home/someone/music%20directory" not "%2Fhome%2Fsomeone%2Fmusic%20directory", so at first I thought that I need to do it manually but after looking at GLib.Uri doc found that they have provided argument for ignoring characters.

Thanks for suggestion. :D

src/LocalBackend/LocalLibrary.vala
mohansharma-me: Think of a situation where you have a 
1. music file with title "English Song" (/home/someone/musics/latest songs/1.mp3) and 
2. a directory "/home/someone/musics/English Songs", 
okay. 

Now when you type "uri:english songs" in searchField then depending on "string[] fields" the result varies.

For example, if "uri" field included with other fields then the result will contain both the paths and if we use only "uri" field (excluding other fields) then our result will only contain that directory path. (which is goal of uri: prefix search).

Let me know if you need to change it to mix both results but that will not help users to create separate result for particular directory.

Thanks :D

mohansharma-me: Primary goal for "uri:" filter is to "list the music from particular directory/path" and "create smartplaylist from uri filter".

:D

matzipan: Sure but there are implementation limitations, not conceptual limitations. For all I can tell, there's nothing limiting an implementation where a query like `uri:"/some/path/here" artist:"Andrei Zisu"` should not be possible.

mohansharma-me: I didn't understand. Can you please clarify what are you saying ?

Am I supposed to remove this filter or what !?

Thanks :D

matzipan: Ok so it's my understanding that you can't use `uri` and `artist` at the same time, right? Why is that?

mohansharma-me: No, I/We can use `uri`, `artist` or `any-field` at same time, It will be simple, but it will mix-up the result. 

Actually I got this idea when I started using noise and then I was searching for the option to filter search result on location only (as most of the time I usually play music directory wise) but didn't found one. So implemented it.

We can also combine different filters in a single query just like you suggested `uri:"/path/to/musics" artist:"Andrei Zisu"`.

The only reason to not include `uri` with other fields is to separate music from that particular path. 

:D

matzipan: Sorry for the long delay - was a bit busy.

Well if you wanna filter by location only - then you just use the URI filter and nothing else.

matzipan: Honestly I don't see any reason why this filter should be singled out from the rest.

mohansharma-me: No worries about delay :)

Okay, I will mix up the result but then in some situation users have to type full path and also we don't need `uri:` prefix now.

I think you're right in a sense because without using `uri:`, it would be simple for end users also.

matzipan: > but then in some situation users have to type full path and also we don't need uri: prefix now.

What do you mean?

src/LocalBackend/LocalLibrary.vala
```diff
@@ -566,19 +566,14 @@ public class Noise.LocalLibrary : Library {
...
-                } else {
-                    fields = {"title", "artist", "composer", "album_artist", "album", "grouping", "comment"};
-                }
+                string[] fields = {"title", "artist", "composer", "album_artist", "album", "grouping", "comment", "uri"};
 
                 foreach (var field in fields) {
                     var id_field = sql.add_id (field);
-                    var id_value = sql.add_expr_value (null, "%"+new_search+"%");
+                    var id_value = sql.add_expr_value (null, "%"+search+"%");
+                    if(strcmp("uri", field) == 0) {
```
matzipan: You can just use field == "uri" in Vala-land. The compiler knows how to treat this.

mohansharma-me: Oh thank you for this also :) I will update it.

src/LocalBackend/LocalLibrary.vala
matzipan: Cool stuff. I'll have a final look and then approve this pull request.

mohansharma-me: Okay :)

Now my mid-term exams are gonna start, bye. ;)

mohansharma-me:
Yes, SmartPlaylistEditor I didn't remove anything just added URI as a new filter.

matzipan:
Then I guess this PR does what it says on the tin. Perhaps this might be a point to improve in the future.

mohansharma-me:
Well, Thanks now I can start fixing issue #67.

And I don't have write access to this repo so someone with write access needs to merge pull request.

matzipan:
Somebody will merge this shortly.

mohansharma-me:
It's not merged yet :((

I want to start working on another new feature around search facility and also want to fix issue 67 but can't because I don't want to create multiple branches for all these codes.

tintou:
I just fixed several code style issues, even if the code of Music is very inconsistent, new code should follow the [elementary code style rules](https://elementary.io/docs/code/reference#code-style)
I'm now waiting for continuous integration to be ready.

mohansharma-me:
Ohh thank you for this. I missed those things. :)

PR #80, opened 2017-10-30
clean up preferences window

Nine-H:
I was watching dan's stream when I noticed a bunch of legacy code from when noise had tabs in the preferences window (luna)

- deleted stack
- deleted notebook switcher
- changed weird layout code to just use a grid
- rewrote the preferences dialog as a single class
- deleted a bunch of single-use methods that really just wanted to be a construct block
- deleted 50 whole lines all up


Nine-H:
@tintou your commit broke my changes. I deleted the stack switching code because it hadn't been used since luna. I tested my version of PreferencesWindow.vala and it worked fine; I'm not sure I understand what happened there.

danrabbit:
@Nine-H it looks like when you started your branch you might not have been branching from the latest git master?

tintou:
@Nine-H I tried to fix the conflict you can see here, but GitHub seems to be unable to fix it

Nine-H:
weird, I'lll cut, pull, paste, push and see if that fixes it.

Nine-H:
that looks better

PR #83, opened 2017-11-01
removed outdated build files

Nine-H:
Noise master currently has a `cmake` directory that's out of date compared to the cmake configs shipped in `elementary-sdk`. It also contains a number of .vapi files that are _waaaaaaaaaaaaay_ out of date. I scrubbed these and that squelched a bunch of deprecated warnings when building.

- deleted outdated cmake config
- deleted out of date .vapi files

I left `libgda-5.0.vapi` in place: I tried and failed compiling a new `libgda-5.0.vapi` from `libgda-5.0.gir`. the remaining deprecated warnings are from libgda. Here's hoping this makes someone set up meson.


Nine-H:
travis didn't like that one bit. I'll look into how other official repos work or try a pull that just rms the vapis.

danrabbit:
pinging @codygarver to see if he would approve of removing the modules from this branch.

In order to make Travis happy, you'll need to add `cmake-elementary` to the packages to pull in. Also make sure to add it as a dependency in the README

tintou:
You need to carefully compare libaccounts-glib.vapi from this repo and libaccounts-glib.vapi from Loki, I'm sure that there are fixes that I introduced there in the ownership, the same goes for libgsignon-glib, the other are safe to go

Nine-H:
alright, travis is happy, I'll diff libaccounts and libgsignon when I get home.

Nine-H:
I restored the vapis in this branch, and filed a bug for some other dude to take a look at them #90

codygarver:
I do not agree that it's a good idea to replace the shipped modules with `cmake-elementary`.

I'm hoping to get rid of the need for the package as soon as possible and adding more things that depend on it does not help.

I would prefer if you simply updated the modules to the latest from [cmake-modules](https://github.com/elementary/cmake-modules) instead.

jeremypw:
@Nine-H :  Are you OK to update the cmake directory as suggested by cody?  It may be cleaner to simply start a new PR.

Nine-H:
I'll give it a shot, it might be more productive to let you get the vapis done first because they're generating the build errors, and do some small patches in the mean time.

Bat41:
I think we can close this because we are switching to Meson anyway.

PR #88, opened 2017-11-20
Better repeat

### LINKED ISSUE #71: Looping over 1 song
I might be missing something, but there is no option that loops over a single song. Is that a bug or something that has not been implemented ?

Bat41:
There is now three possible states accessible by clicking on the repeat button: none, all, or song. Really fixes #71


Bat41:
Thanks for your review! But I'm not sure I understand what you said there:

> In master, when you choose an option from the menu that option is persistent, but in this branch, after switching off repeat then it goes back to "repeat song". I feel this is a regression.

Or at least, I can't reproduce.

jeremypw:
@Bat41: Sorry to be unclear.  This is what I mean:

* Start with repeat off
* Secondary click and set to "Album"; button (tooltip ) switches to "repeat album" mode
* Continue primary clicking; master toggles between "Off" and "Album" modes. But PR requires two clicks to turn off and then goes back to "Song" and then "All";  "Album" mode is lost (not persistent).

Actually I now realise that the tooltips in master are also inconsistent - Some indicate future state ("Enable repeat" and "Disable repeat" while others indicate current state "Repeat Song" etc) so it is not a fault particularly with this PR.

Nine-H:
could we create a playback settings heading in the preferences window? with repeat mode and shuffle mode settings so we don't need to do weird "click twice to x, click hold and spin around three times to y" context menus on buttons got removed from the UI in freya and it was a massive improvement.

Bat41:
I updated this branch according to what have been said in #117, I think it's ready for review (and I take the discussion we had in the issue as the opinion of the UX team, so I remove the "Needs design" label).

Bat41:
OK, I don't know what I did with git there. I'll start a new clean branch.

PR #89, opened 2017-11-25
Read the embedded cover art

### LINKED ISSUE #47: Music doesn't read embedded Cover art
I've seen this with my library of music, it doesn't pick up that the files have embedded cover art in them, instead, it needs a cover.jpg file to be able to show cover art, which isn't normal behavior.

Bat41:
For some reason, the async API of `Gst.PbUtils.Discoverer` was never emitting the `discovered` signal. I changed the code to use the sync API, but in a new thread, to avoid freezing the UI.

Fixes #47


src/GStreamer/CoverImport.vala
```diff
@@ -35,24 +35,25 @@ public class Noise.CoverImport : GLib.Object {
...
-            discoverer.discover_uri_async (media.uri);
-        }
-    }
-
-    public void start () {
-        discoverer.start ();
+        new Thread<void*>(null, () => {
+            lock (this.album) {
+                foreach (var media in album.get_media ()) {
+                    var info = discoverer.discover_uri (media.uri);
```
danrabbit: There's a warning here in Terminal about an unhandled GLib.Error

Bat41:
Should be OK now.

PR #115, opened 2017-12-02
Remove/Rename redundant icons

danrabbit:
Remove icons that are covered under the standard freekdesktop icon naming spec or are provided by both Adwaita and elementary

Fix calls to icon names that have standard name counterparts


danrabbit:
ty ty

PR #120, opened 2017-12-02
Remove legacy dialog

### LINKED ISSUE #24: Rewrite SyncWarningDialog as Gtk.Dialog
Dialogs should be a Gtk.Dialog subclass, not Gtk.Window

Launchpad Details: [#LP1310480](https://bugs.launchpad.net/bugs/1310480) Daniel Fore - 2014-04-21 07:03:44 +0000

Bat41:
Closes #24


danrabbit:
I don't think we want to remove the ability to import files from an external device

Bat41:
This code was never called.

PR #135, opened 2017-12-03
Granite SeekBar

### LINKED ISSUE #110: Use SeekBar from Granite
what's on the tin

Bat41:
Fixes #110


Nine-H:
love it, was just about to do this. made a small edit to stop the label scraping the top of the window, just nixes some row padding. keep it up boys.

PR #136, opened 2017-12-03
Built-in MPRIS

### LINKED ISSUE #108: Built in MPRIS
MPRIS should be supported without any plugin.

Bat41:
Fixes #108 

Before testing, make sure to uninstall the old MPRIS plugin or to disable it by setting `org.pantheon.noise.settings.plugins-disabled` to `['mpris']`. If you don't, some types will be registered two times, and the app will just crash. I don't know if the plugin will be uninstalled when users will update.


Nine-H:
build failed
`fatal error: libindicate/listener.h: No such file or directory`
I have `libindicate-dev` and `libindicate-gtk3-dev` installed, please advise.

Thalexander:
dpkg shows that file as owned by libindicate-dev so it should work if you have it installed, maybe try a restart to see if it updates the path.
libindicate-dev: /usr/include/libindicate-0.7/libindicate/listener.h

tintou:
It would probably better to rewrite the MPRIS support without libindicate and Dbusmenu-0.4, it seems like no one is using those libraries directly

tintou:
Noise.MPRIS should probably just be a namespace containing two static functions, otherwise it looks very good

tintou:
Looks great!

PR #137, opened 2017-12-03
cleaned up musiclistview.vala

Nine-H:
doesn't do anything ~~~fun~~~ controversial. just changes a bunch of method names and gets rid of a `build_ui()` begs the question why a context menu is called "MusicListView".


src/Views/ListView/Lists/MusicListView.vala
```diff
@@ -84,109 +82,73 @@ public class Noise.MusicListView : GenericList {
         set_compare_func (view_compare_func);
 
         // Don't reorder the queue
-        /*if (playlist == App.player.queue_playlist) {
-            set_sort_column_id (-2, Gtk.SortType.DESCENDING);
+        /*if (playlist != App.player.queue_playlist) {
```
Bat41: In fact I think you can even delete this comment…

src/Views/ListView/Lists/MusicListView.vala
```diff
@@ -84,109 +82,73 @@ public class Noise.MusicListView : GenericList {
...
+        media_rate_media = new Granite.Widgets.RatingMenuItem ();
+        media_rate_media.activate.connect(media_rate_media_clicked);
+
+        media_action_menu = new Gtk.Menu ();
+        media_action_menu.attach_to_widget (this, null);
         if(hint != ViewWrapper.Hint.ALBUM_LIST) {
-            //mediaActionMenu.append(browseSame);
-            mediaActionMenu.append(mediaScrollToCurrent);
-            mediaActionMenu.append(mediaScrollToCurrentSeparator);
+            //media_action_menu.append(browseSame);
```
Bat41: Here too, you can just delete this line.

src/Views/ListView/Lists/MusicListView.vala
```diff
@@ -84,109 +82,73 @@ public class Noise.MusicListView : GenericList {
...
+            media_action_menu.append(media_scroll_to_current);
```
Bat41: It would be perfect if you made sure every opening parenthesis is preceded by a space, but if you don't want to check the whole file, just tell and I'll make a little commit.

Bat41:
> begs the question why a context menu is called "MusicListView".

Well, it's actually a list view : it inherits `GenericList`, but it mostly hads a context menu to this list, yes. We should maybe check if there is any other `GenericList` subclass, maybe we could merge them…

Nine-H:
alright lets land this

PR #141, opened 2017-12-03
PopupListView.vala: Remove window resizing

danrabbit:
Allow the window to be resized, don't try to resize it


Bat41:
Not sure of what you want to do. Why do you want to make this window resizable?

danrabbit:
Kind of reworking this class so it could be used as a sidebar, but also I don't see a good reason for this dialog to not be resizable. It seems kind of overengineered to try to force the window to be square, imo

PR #145, opened 2017-12-03
Album popup is now album sidebar

danrabbit:
screenshot from 2017-12-03 10 27 23


Bat41:
A screenshot where the padding issue is visible:

capture d ecran du 2017-12-03 19 32 54

danrabbit:
Fixed the wrapping issue and renamed file since it's no longer a dialog.

In my experience, Gtk.Revealer doesn't play nice with Gtk.Paned. Can we save that for another branch?

PR #149, opened 2017-12-03
CSS album art fallback

danrabbit:
* Fixes albumart size for HiDPI
* Uses `card` class for shadows
* Draw the fallback icon from the system icon set
* Set a sane minimum width for the album art sidebar

screenshot from 2017-12-03 20 31 13


danrabbit:
Got verbal approval from @tintou

PR #150, opened 2017-12-03
Merge Noise.vala and main.vala into Application.vala

danrabbit:
Because that's how we do it in other apps


Nine-H:
it's kind of a shame it's pulling in stuff from the build class (instead of setting these properties with Object()) and calls a build_ui() method, but that's a problem with LibraryWindow and noise in general not consolidating init, also testing file opening uncovered a bug in master so, cool. approved.

PR #153, opened 2017-12-04
Remove star ratings

### LINKED ISSUE #86: Ratings don't work
#### Steps to Reproduce

1. Ensure you are in list view
2. Display the rating column
3. Right click on a track
4. Select a rating (stars)

#### Expected Result
The rating is displayed in the respective column

#### Actual Result
The rating doesn't change

#### Comments
I've tried various other ways to rate tracks such as directly clicking into the rating column or using the Edit Song Info dialog. None of them seem to have any effect.

### LINKED ISSUE #107: Remove Star Ratings
Star ratings are dumb

danrabbit:
Fixes #86 
Fixes #107 

Removes ratings, which are currently busted anyways


danrabbit:
I'm gonna put a block on this branch, just because I think it will require some regression testing, especially with smart playlists

danrabbit:
There was some discussion in slack about smart playlists and what should happen to playlists based on ratings. This branch removes the ability to configure ratings in smart playlists, but it doesn't actually remove the rule from existing smart playlists, so they still filter based on ratings. This could be a problem since there's no way to tell that a playlist is being filtered based on ratings and no way to change that behavior

PR #154, opened 2017-12-04
LibraryWindow.vala: misc cleanup

danrabbit:
* Reduce variable scope
* Remove commented code
* Remove unneccesary comments
* Combine method that is only used once
* Space before ()


Nine-H:
it's heaps easier to understand UI construction now. still some absolute clangers like that if chain in playlist_name_edited() but good stuff.

src/LibraryWindow.vala
```diff
@@ -983,9 +945,6 @@ public class Noise.LibraryWindow : LibraryWindowInterface, Gtk.Window {
         // if not 90% done, skip it
         if(!added_to_play_count) {
```
Bat41: Maybe check for spaces before parenthesis too?

PR #155, opened 2017-12-04
LibraryWindow.vala: fix JS-style variable names

danrabbit:
Also fix code style for lines we touch


Nine-H:
not entirely certain what the difference is between this and the other branch with LibraryWindow, but it's 1am, and it builds, and I can't see anything wrong with this (DRY issues notwithstanding), so, all good.

PR #160, opened 2017-12-04
Remove UI class

Thalexander:
Not used anywhere.


Nine-H:
good pick, I like dead code removal.

PR #164, opened 2017-12-04
Add similar playlist icon to gresource

danrabbit:
instead of installing it to the filesystem


Bat41:
Is creating one resource file by plugin too much, or can we separate things as much as possible?

tintou:
@Bat41 It's a good question, I thought about it but one answer is that some other plugin might want to reuse this icon so it's maybe better to ship in in the app itself

PR #168, opened 2017-12-04
AlbumListGrid.vala: misc cleanup

danrabbit:
* Explicitly mark private variables as private
* Reduce scope of cover_set_new
* Make sure that we update the album image if we set a new one from the file chooser
* Pass parent window to file chooser


src/Widgets/AlbumListGrid.vala
```diff
@@ -39,18 +39,14 @@ public class Noise.AlbumListGrid : Gtk.Grid {
...
     }
 
-    Widgets.AlbumImage album_cover;
-    Gtk.Label album_label;
-    Gtk.Label artist_label;
-
-    Gtk.Menu cover_action_menu;
-    Gtk.MenuItem cover_set_new;
-
-    Granite.Widgets.Rating rating;
```
Bat41: Does removing this belongs there or to #153?

danrabbit: It's just moved down here as `private Granite.Widgets.Rating rating;` on line 49 :)

Bat41: Oh, sorry, I didn't saw that. :grin:

PR #171, opened 2017-12-04
halign prefs grid to avoid fucky switches

### LINKED ISSUE #176: Preference dialog switches stretch with a long bookmark folder
If you have a long folder bookmarked, it expands the file chooser dropdown which expands the whole dialog, switches included.

pasted image at 2017_12_04 04_23 pm

cassidyjames:
Fixes #176 

## Before
pasted image at 2017_12_04 04_23 pm

## After 
screenshot from 2017-12-04 16 35 59


danrabbit:
Hm I think this is a symptom of a larger problem. This is what I have:

screenshot from 2017-12-04 13 13 32

cassidyjames:
@danrabbit strange. I was on LoDPI here, let me look at both that and HiDPI.

cassidyjames:
Not sure what caused it in the first place, but it looks like it's reproducible with #176. Updated this branch to halign the switches to fix it.

PR #174, opened 2017-12-04
Allow removing media from album view

danrabbit:
While we're here, use switch/case for legibility


danrabbit:
Superseded by #183

PR #175, opened 2017-12-04
AlbumListGrid.vala: Show album cover menu on left click too

danrabbit:
A bit more discoverable if you can do left or right click here


Nine-H:
this is cool but in testing I noticed that the art in the album sidebar doesn't load when it's set, just in the album grid.

danrabbit:
That also occurs in trunk. This branch doesn’t seek to resolve that issue

PR #195, opened 2017-12-07
ViewSelector.vala: misc cleanup

danrabbit:
Replace get/set methods with properties


Nine-H:
I saw this needs review, the code looks good, but when I checked the branch out and built it I couldn't change views at all.

danrabbit:
I split this branch up into a view smaller ones to try to find out where the regression is happening

PR #198, opened 2017-12-07
LibraryWindow.vala: port saved_state to GLib.Settings

danrabbit:
Port the saved state settings in LibraryWindow to GLib.Settings instead of Granite Settings

* Remove now unused properties and enum from Settings.vala
* Decrease the scope of `window_height` and `window_width` to `configure_event`
* Increase the scope of `saved_state_settings` instead of recreating it in several methods
* Respect window manager defaults for placing windows without saved coords (For Gala this is still center, so no change)
* Bind `sidebar-width` instead of get/set

### DIFF
--- a/core/Settings.vala
+++ b/core/Settings.vala
@@ -28,11 +28,6 @@
 
 namespace Noise.Settings {
     public class SavedState : Granite.Services.Settings {
-
-        public int window_width { get; set; }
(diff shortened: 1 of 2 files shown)


Nine-H:
opening and closing the window without changing position and sidebar width makes the app hang for a while. not sure why.

PR #200, opened 2017-12-08
ViewSelector.vala: Adjust modebutton properties

danrabbit:
* Organized so that every time we touch modebutton is in the same place
* Remove unnecessary set size request
* Removed unnecessary halign and valign
* Removed unnecessary invisible widget. Set selection to -1 instead
* Minor code style tweaks, brackets, spacing, etc


Nine-H:
did some tests, if the app starts in album grid and you click column browser then the mode switches to list view. I'll take a look at the logic after work.

PR #206, opened 2017-12-10
Save / restore window position

### LINKED ISSUE #184: Save/Restore window position
Looks like we save size, but not position

mdh34:
Music now saves and restores window position, fixes #184

### DIFF
--- a/src/LibraryWindow.vala
+++ b/src/LibraryWindow.vala
@@ -439,7 +439,7 @@ public class Noise.LibraryWindow : LibraryWindowInterface, Gtk.Window {
         var window_x = saved_state_settings.get_int ("window-x");
         var window_y = saved_state_settings.get_int ("window-y");
         if (window_x != -1 ||  window_y != -1) {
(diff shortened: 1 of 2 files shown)


src/LibraryWindow.vala
```diff
@@ -436,7 +436,7 @@ public class Noise.LibraryWindow : LibraryWindowInterface, Gtk.Window {
         saved_state_settings = new GLib.Settings ("org.pantheon.noise.saved-state");
 
         set_default_size (saved_state_settings.get_int ("window-width"), saved_state_settings.get_int ("window-height"));
-
+        move (saved_state_settings.get_int ("window-x"), saved_state_settings.get_int ("window-y"));
```
danrabbit: You should check that the values are not `-1, -1`. Otherwise the window will be placed at `0, 0` instead of the window manager default (center) :)

mdh34:
Just fixed!

danrabbit:
Awesome, thanks! Approved :D

PR #209, opened 2017-12-10
Statusbar menubutton sensitive

danrabbit:
We already run this check (`have_media`) in LibraryWindow.vala in order to set the source list category visibility. Instead of running the same check twice, just use the cached result of the first check and make statusbar dumber

### DIFF
--- a/src/LibraryWindow.vala
+++ b/src/LibraryWindow.vala
@@ -643,7 +643,7 @@ public class Noise.LibraryWindow : LibraryWindowInterface, Gtk.Window {
 
         // hide playlists when media list is empty
         source_list_view.change_playlist_category_visibility (have_media);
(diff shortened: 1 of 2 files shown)


src/LibraryWindow.vala
```diff
@@ -643,11 +643,10 @@ public class Noise.LibraryWindow : LibraryWindowInterface, Gtk.Window {
 
         // hide playlists when media list is empty
         source_list_view.change_playlist_category_visibility (have_media);
+        statusbar.playlist_menubutton_sensitive = have_media;
```
davidmhewitt: For completeness, should this not be:
```statusbar.playlist_menubutton_sensitive = folder_set && have_media;```

Just to make it the same as the check you're moving.

PR #214, opened 2017-12-16
Correctly pluralize media/medium

Bat41:
I probably forgot things, but it's a good start.


Nine-H:
I'm not sure this change is appropriate, because "media" is short for "multimedia" hence the constant pluralisation.

tintou:
@Nine-H Hmm, I'm pretty sure that Multimedia is composed of `multi` meaning several and `media` (plural of medium)

Nine-H:
medium sounds very very wrong to me as a native english speaker. To drag it back to the 90s the media would be the albums and singles on the shelf and the medium would be the chain of record stores, the mail order form in the magazine, or if you really like dad rock: kmart.

In a more modern setting the media would be the files or streams and the medium would be internet radio, ondemand streaming platforms, youtube, torrent tracker, mp3 player etc.

danrabbit:
Yeah I’m gonna agree with @Nine-H here. Medium seems very odd in this context

Bat41:
Okay, let's close that.

PR #219, opened 2017-12-20
LibraryWindow.vala: Make public things private

danrabbit:
Several methods were marked as public but only used in this class

`media_not_found` was never used at all

### DIFF
--- a/src/LibraryWindow.vala
+++ b/src/LibraryWindow.vala
@@ -273,7 +273,7 @@ public class Noise.LibraryWindow : LibraryWindowInterface, Gtk.Window {
         connect_to_sourcelist_signals ();
     }
 
-    public void connect_to_sourcelist_signals () {
+    private void connect_to_sourcelist_signals () {
 
         source_list_view.selection_changed.connect ((page_number) => {
             view_container.set_current_view_from_index (page_number);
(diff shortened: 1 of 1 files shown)


src/LibraryWindow.vala
```diff
@@ -1094,15 +1094,7 @@ public class Noise.LibraryWindow : LibraryWindowInterface, Gtk.Window {
         }
     }
 
-    public void media_not_found(int id) {
-// XXX FIXME TODO Don't depend on ids
-#if 0
-        var not_found = new FileNotFoundDialog(library_manager, this, id);
```
Bat41: Is `FileNotFoundDialog` used somewhere else?

danrabbit: Yes, it's called from `Views/ListView/Lists/MusicListView.vala`

Nine-H: I couldn't get the file not found dialog to pop up because after deleting a track from my disk noise whipped it out of my library on startup, when I deleted one while in operation the app removed it from the library when I played it (in master). I think if musics going to be smarter and more automatic we should assume whatever we can't find is media that doesn't exist. However: because this is a small commit centered around removing access to class methods and doesn't introduce any regressions I give it my approval.

PR #224, opened 2017-12-22
Mediakey code style

danrabbit:
* Use brackets
* Space before `()`
* cuddle `catch`
* Use a switch/case instead of elseif/elseif/elseif


src/Objects/MediaKeyListener.vala
```diff
@@ -54,45 +55,43 @@ public class Noise.MediaKeyListener : Object {
...
-            }
-            catch(IOError err) {
+            } catch (IOError err) {
                 warning ("Could not grab media player keys: %s", err.message);
             }
         }
     }
-    
+
     public void releaseMediaKeys() {
```
Bat41: Maybe change the method's names to `snake_case` too?

PR #232, opened 2017-12-27
InstallGstreamerPluginsDialog.vala: Rewrite, fix deprecations

danrabbit:
For some reason this is throwing a fit when trying to subclass `Granite.MessageDialog`, but at least this:

* Fixes deprecations with `margin_left` and `margin_right`
* Use two labels for primary and secondary text
* Use primary CSS class from Granite
* Correctly align and wrap labels
* Use suggested_action class for install button


jeremypw:
If you wish to base this on Granite.MessageDialog, you must call the base constructor with the appropriate parameters e.g.

```
 public InstallGstreamerPluginsDialog (Gst.Message message) {
        base.with_image_from_icon_name ("primary text", "secondary text", "dialog-question");
        ........
```

You can the discard some of the current code.

jeremypw:
Is there an easy way to trigger the use of this dialog?

danrabbit:
@jeremypw Thanks for the tip! I have it subclassing Granite.MessageDialog now

Uninstall `gstreamer-plugins-good` and try to play an MP3 :)

PR #236, opened 2018-01-04
Add data tests

danrabbit:
Add appstream and desktop validation tests


danrabbit:
Something is wrong here. The tests can't find the files in Travis, but it seems to work locally

Bat41:
It works fine on my own computer too. Should we merge this (and try to fix Travis in another PR) or try to understand what is wrong with it?

danrabbit:
It would be better to fix it first. Otherwise all other merges will be blocked by failing tests

PR #240, opened 2018-01-06
RDNN project name

danrabbit:
Remove some unused CMake variables while we're here

I guess there's an open question about what the core library name should be. At the moment it becomes `io.elementary.music-core`


CMakeLists.txt
```diff
@@ -2,20 +2,17 @@
...
 enable_testing()
 
-set (GETTEXT_PACKAGE ${CMAKE_PROJECT_NAME})
+set (GETTEXT_PACKAGE "noise")
 set (RELEASE_NAME "Luna")
 set (VERSION "0.4.2")
 set (VERSION_INFO "Stable Release")
 
 include (GNUInstallDirs)
-set (DATADIR ${CMAKE_INSTALL_PREFIX}/share)
```
Bat41: If you remove this line, translations are no longer loaded. Or if you really want to remove it you can replace `string langpack_dir = Path.build_filename (Build.DATADIR, "locale");` (line 114, in `Application.vala`) with `string langpack_dir = Path.build_filename ("/usr", "share", "locale");`

PR #246, opened 2018-01-11
Misc deprecation fixes

danrabbit:
* `margin_left` and `margin_right` have been replaced by `margin_start` and `margin_end`
* `set_rules_hint` isn't a thing anymore. You can't have alternating colored rows in treeview :(
* `follow_state` isn't a thing anymore because according to docs cell renderers always follow state
* `Gtk.Combobox.set_title` isn't a thing anymore because this was related to "tear off mode" whatever that is


tintou:
Please keep set_rules_hint It's not deprecated but is a regression in Gtk itself

danrabbit:
@tintou naw it is actually deprecated. Look:
```
/src/Views/ListView/Lists/GenericList.vala:80.9-80.22: warning: Gtk.TreeView.set_rules_hint has been deprecated since 3.14
```

PR #248, opened 2018-01-11
No more video playback

### LINKED ISSUE #242: Remove Vestigial Video Playback
For some reason noise can play back videos. This is silly.

Nine-H:
This really does fix #242 I couldn't null out the video sink in our playbin, but I could substitute in a fakesink. If this is totally bonkers then feel free to delete this pull request. technically you _can_ playback a video, but no playback window opens and you can't get into an undefined state by closing the window.

### DIFF
--- a/core/GStreamer/Pipeline.vala
+++ b/core/GStreamer/Pipeline.vala
@@ -37,6 +37,8 @@ public class Noise.Pipeline : GLib.Object {
     public dynamic Gst.Element audiosinkqueue;
     public dynamic Gst.Element eq_audioconvert;
     public dynamic Gst.Element eq_audioconvert2;
+    
+    public dynamic Gst.Element fake;
 
     public dynamic Gst.Element playbin;
     public dynamic Gst.Element audiotee;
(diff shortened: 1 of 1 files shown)


danrabbit:
Eh it seems weird to add code in to prevent playback. If it’s something that just happens for free it seems like it’s easy to just ignore it

Nine-H:
there's a few weird behaviours you can trigger by opening a video file, the worst one is that closing the window puts you in an undefined state between pause and play that breaks the playback controls. Perhaps blocking video files from being opened is a better solution but I'm not sure how to achieve that.

core/GStreamer/Pipeline.vala
```diff
@@ -47,6 +49,9 @@ public class Noise.Pipeline : GLib.Object {
 
         pipe = new Gst.Pipeline("pipeline");
         playbin = Gst.ElementFactory.make ("playbin", "play");
+        
+        fake = Gst.ElementFactory.make ("fakesink", "fake");
+        playbin.set ("video-sink", fake );
```
Bat41: `playbin["video-sink"] = fake` would be better (or even `playbin.video_sink = fake`, but not sure it will work).

danrabbit:
Can we just open videos with XDG-Open instead?

Nine-H:
In an ideal world noise would reject videos and we wouldn't have any video decoders or sinks in our pipeline. If the player is going to be internet facing (for streaming) we'd want to ristrict the gstreamer plugins that can actually be loaded because some (particularly gstreamer-plugins-bad and gstreamer-plugins-ugly) have had exploits before.

danrabbit:
Gonna close this since @Bat41's branch will just open Videos which seems to not break functionality while giving us a better video player :+1:

PR #261, opened 2018-01-14
Port playback controls to GLib.Action

danrabbit:
This branch ports the play/pause, next, and previous functions to GLib.Action. The reason I want to use GLib.Action here is because it allows us to reduce the amount of functions that are touching the UI. This makes it easier for us to move the whole headerbar into its own class in the future, which I think would be nice to break up this huge LibraryWindow class a little.

Since we don't have to touch the actual UI anymore when setting the playback action state, we can actually set this directly from the playback manager instead of having to listen for a signal in LibraryWindow.

### DIFF
--- a/src/LibraryWindow.vala
+++ b/src/LibraryWindow.vala
@@ -48,7 +48,6 @@ public class Noise.LibraryWindow : LibraryWindowInterface, Gtk.Window {
     private bool media_half_played_sended { get; set; default = false; }
     private bool search_field_has_focus { get; set; default = true; }
 
-    private Gtk.Button play_button;
     private Gtk.Paned main_hpaned;
     private Cancellable notification_cancellable;
     private PreferencesWindow? preferences = null;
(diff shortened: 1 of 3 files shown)


src/LibraryWindow.vala
```diff
@@ -1010,6 +1024,18 @@ public class Noise.LibraryWindow : LibraryWindowInterface, Gtk.Window {
         }
     }
 
+    private void action_play () {
+        play_media ();
```
Bat41: Can't we just give this method when defining the action?

danrabbit: It will complains because these methods take an argument. That's something I think I'd like to address in a later commit is removing this "inhibit_notifications" argument since we should probably just be deciding that based on the focus state of the window. I need to look into what else calls this method though and figure out what cases they're inhibiting notifications under and if there's something I'm missing

PR #262, opened 2018-01-14
Headerbar

danrabbit:


Nine-H:
This works, and the branch has been updated to work with meson is there any reason why it hasn't been committed yet?

danrabbit:
Well I kind of did it as an experiment, but so far I don't know that it really makes the code better

BaptisteGelez:
If we realize that it should be moved to the core, it will never be too late to open another PR to change that.

Issue #260, opened 2018-01-14
Plugins are depending on the app


elegaanz:
Plugins should only depend on the core library, but they need the app files to build.

danrabbit:
Thanks for your report! We're doing a from-scratch rewrite of Music and this issue no longer applies to the new code base. I'm closing this report since the issue can't be reproduced in the new version of Music.

Issue #207, opened 2017-12-10
Make column view an actual separate view


danrabbit:
And then create an instance of listview inside of it like we do with album view.

This would simplify view switching quite a bit

danrabbit:
Thanks for your report! We're doing a from-scratch rewrite of Music and this issue no longer applies to the new code base. I'm closing this report since the issue can't be reproduced in the new version of Music.

Issue #186, opened 2017-12-05
Selecting shuffle chooser resets view to list view [$50]


danrabbit:
Every time you change shuffle mode you get kicked to list view and the view chooser becomes insensitive

---
There is a bounty on this issue in [Bountysource](https://www.bountysource.com/issues/52352815-selecting-shuffle-chooser-resets-view-to-list-view)

Bat41:
I can reproduce, but only if music is playing.

Nine-H:
This seems related to search and the modebutton being set insensitive. There's also some weird shit going on in `src/Widgets/ViewSelector.vala`

danrabbit:
It seems like this can be reproduced by changing the queue in any way

Issue #147, opened 2017-12-03
Changing song opens the list view


BaptisteGelez:
Steps to reproduce:

- open an album in the sidebar
- double click on any song: it plays
- double click on another song: it plays, but you now are in the list view

It also happens when you are in column view.

PR #265, opened 2018-02-03
Rewrite view system

BaptisteGelez:
I started a complete rewrite of the view management.

With this branch, we now have a central place for managing views : the `ViewManager`. It will send signals when a view is added or removed, so that the UI can be updated. This class is defined in the core (and in the future, its only instance will be living there, but for now I put it in the main window).

Views are subclasses of `Noise.View` : they have a title, an ID and an icon. They are `Gtk.Bin`, that can contain whatever you want.

One of my goal with this branch is also to get rid of the `ViewWrappers`, which seems overcomplicated and quite useless. I'm also trying to make view management more abstract and less couples to the app, so that it will be easier for plugins to add their views.

~~I had to comment a lot of things to make it compile, but I will reintroduce all these features when the base will be ready.~~

~~There is still a lot of work to do : I only implemented a basic ViewManager, that I'm using to populate the left sidebar. But I guess once it will be ready, it may fix all the bugs we have with the current view system.~~


tintou:
You can even remove the Noise.View class, using Gtk.Stack + child properties should be enough for this and is the "right" way to do it with Gtk+

BaptisteGelez:
@tintou Well, I added other properties in my last commit, so maybe this class is still needed. And I'm not sure if it would work, because we may need to access these properties before the views are added to the stack.

BaptisteGelez:
A list of issues this PR should close:

- #207 
- #186 
- #147 
- #92 
- #18

And it is a first step to achieve #260 too

BaptisteGelez:
I think this PR is ready to be reviewed and merged. But please, if you merge it, keep each commit independent (don't squash them) so we can `git bissect` easily if a regression is found after the merge, since it's a quite big change.

BaptisteGelez:
Thanks for your review!

> I noticed there's a few things here that can be broken out like renaming variables

When I did I think it was because it was necessary (for instance, a `title` widget was conflicting with the `title` property of the view), or at least in a part of the code I had to change anyway.

> changing some dialog properties in SmartPlaylistEditor.vala

They were set by the parent view, in a method I removed.

For code style and debug messages changes, I'll revert them.

danrabbit:
Yeah that's fine, I understand why you made those changes. I just mean that they could be done in a bunch of smaller branches instead of one giant branch. This would make those individual changes much easier to review. A branch this big is really difficult to review

src/Views/AlbumsView.vala
```diff
@@ -23,38 +22,42 @@
  * obligated to do so. If you do not wish to do so, delete this exception
  * statement from your version.
  *
- * Authored by: Scott Ringwelski <sgringwe@mtu.edu>
- *              Victor Eduardo <victoreduardm@gmail.com>
+ * Authored by: Baptiste Gelez <baptiste@gelez.xyz>
```
danrabbit: Can you only append your author info here instead of removing previous authors?

core/Views/ViewManager.vala
```diff
@@ -0,0 +1,213 @@
...
+        view_removed (view);
+    }
+
+    /**
+    * Try to retrieve a registered view.
+    *
+    * @return The requested view if found
+    * @throw Throws an error if the view couldn't be found
+    */
+    public new View get (string id) throws ViewError {
```
tintou: This is not good, don't use the `new` attribute, it's hiding things and leads to mistakes, better rename that function

src/Views/DeviceMusicView.vala
```diff
@@ -0,0 +1,125 @@
...
+
+    private void sync_finished () {
+        list_view.set_media (library.get_medias ());
+    }
+
+    public override bool filter (string search) {
+        return true;
+    }
+
+    public override Gtk.Menu? get_sidebar_context_menu (Granite.Widgets.SourceList list, Granite.Widgets.SourceList.Item item) {
```
danrabbit: I'm not a big fan of using a `get_foo` method for building a widget. If there's a change for abstraction here, then I think it makes more sense to create a class

Nine-H:
there's two major blocking regressions: `ListColumns` not being saved and restored correctly in `MusicListViews` and a lack of correct sorting behaviour in any music list.

src/Views/ListView/ColumnBrowser/ColumnBrowser.vala
```diff
@@ -209,7 +206,6 @@ public abstract class Noise.ColumnBrowser : Gtk.Grid {
 
         column_chooser_menu.append (column.menu_item);
 
-        column.row_activated.connect (column_row_activated);
```
danrabbit: Hm this looks like a feature regression :)

danrabbit:
Corentin pointed out that all these replacements of the parent_wrapper library with the main_window library breaks playlists on devices

src/Views/ListView/Lists/TreeViewSetup.vala
```diff
@@ -36,29 +36,55 @@ public class Noise.TreeViewSetup : Object {
...
 
-    public TreeViewSetup (ViewWrapper.Hint hint, string? uid = null, Gda.Connection? connection = null) {
-        Object (
-            connection: connection,
-            hint: hint,
-            uid: uid
-        );
+    public TreeViewSetup (bool for_playlist = false, string? uid = null, Gda.Connection? connection = null) {
+        Object (uid: uid, connection: connection);
+        append_new_column (ListColumn.ICON);
```
danrabbit: I'm not sure this really makes the code better. As I understand it, in an ideal world a constructor only contains calls to `Object ()` which would mean we still need something like `ViewWrapper.Hint hint` internally here even if we don't expose it externally.

src/Views/ListView/ColumnBrowser/ColumnBrowser.vala
BaptisteGelez: Re-introduced this feature

BaptisteGelez:
This PR is too big to be merged, I close it. But I will make smaller PRs to transition from the current system to something similar to this one.

PR #308, opened 2018-03-03
GenericList: Replace parent wrapper with calls to main window

danrabbit:
Some more ports from #265 

Replaces parent_wrapper's library with the main window's library


tintou:
See #309

PR #309, opened 2018-03-03
GenericListView: Use library_manager from main window

danrabbit:
Another port from #265 

Replaces parent_wrapper's library with the main window's library


tintou:
It's a real regression this shouldn't be done.

PR #313, opened 2018-03-20
Gtk.Infobar

Nine-H:
Tooling around with splitting the import progress view from the seek bar. Incomplete and should probably be moved over to extend the view system branch.


danrabbit:
Closing the PR since it hasn't moved in 3 months. Feel free to reopen if work continues :)

PR #321, opened 2018-03-30
Single AlertView

### LINKED ISSUE #252: Add a no search results view
At the moment each view is creating its own ViewTextOverlay instance to display text when no search results are found.

Instead we should have just a single alert view subclass that we switch to like this:

screenshot from 2018-01-11 13 28 26

BaptisteGelez:
Fixes #252


src/Views/ViewStack.vala
```diff
@@ -39,6 +41,9 @@ public class Noise.ViewStack : Gtk.Stack {
                 update_visible ();
             }
         });
+
+        var alert_view = new Granite.Widgets.AlertView ("No results", "Try another search", "edit-find-symbolic");
```
danrabbit: Make sure that these strings are translatable :) Titles should be in title case and descriptions should be sentence case with terminating punctuation. In Switchboard, for example, the strings are:

_("No Results"),
_("Try changing search terms.")

BaptisteGelez:
I can't reproduce your problem when search bar is cleared, it goes back to the normal view for me.

danrabbit:
@BaptisteGelez I can reproduce it in every view:

screenshot from 2018-04-15 09 24 27 2x

src/LibraryWindow.vala
```diff
@@ -298,6 +298,19 @@ public class Noise.LibraryWindow : LibraryWindowInterface, Gtk.Window {
             }
         });
 
+        library_manager.search_finished.connect (() => {
+            print ("search endedn\n");
```
danrabbit: Probably want to either change this to a `debug ()` or remove it before merging :)

danrabbit:
Works as expected now! Left an inline comment, but I think we probably either want to change those `print`s to either `debug` or remove them before merging :)

src/LibraryWindow.vala
BaptisteGelez: Oops, of course. :grin:

PR #331, opened 2018-05-04
Only show alertview when search is non-null

danrabbit:
There's another case where results are <1 and that's when the library is empty. Make sure that there are both no results and we're actually searching so that the welcome view shows on first run instead of the no results view


danrabbit:
Hrm no that's not intended. Is it okay to address that in another branch if it also occurs in master?

PR #339, opened 2018-06-16
Move some keys from settings to saved state

danrabbit:
* Move last media, playback position, and search string from "settings" to "saved state" schema
* Cache results that we use more than once


src/LibraryWindow.vala
```diff
@@ -1152,19 +1158,15 @@ public class Noise.LibraryWindow : LibraryWindowInterface, Gtk.Window {
...
             }
-        }
-        App.player.player.pause ();
 
-        // Search
-        if (!main_settings.privacy_mode_enabled ()) {
-            main_settings.search_string = search_entry.text;
+            App.saved_state.set_string ("search-string", search_entry.text);
         }
+        App.player.player.pause ();
```
jeremypw: Shouldnt this be before you read the position ?  Then, do you need to read the position twice when quitting?

src/LibraryWindow.vala
```diff
@@ -1158,15 +1158,15 @@ public class Noise.LibraryWindow : LibraryWindowInterface, Gtk.Window {
...
+            App.saved_state.set_int ("last-media-position", (int)((double)App.player.player.get_position()/TimeUtils.NANO_INV));
             if (App.player.current_media != null) {
                 App.player.current_media.resume_pos = (int)((double)App.player.player.get_position ()/TimeUtils.NANO_INV);
                 library_manager.update_media (App.player.current_media, false, false);
             }
 
-            App.settings.set_string ("search-string", search_entry.text);
+            App.saved_state.set_string ("search-string", search_entry.text);
         }
+        App.player.player.pause ();
```
danrabbit: Oops accidentally added an extra pause in the merge from master

PR #348, opened 2018-06-17
RemoveFilesDialog: Rewrite as a Granite.MessageDialog

danrabbit:
While we're here, use ngettext for the primary label as well

### DIFF
--- a/src/Dialogs/RemoveFilesDialog.vala
+++ b/src/Dialogs/RemoveFilesDialog.vala
@@ -26,60 +26,41 @@
  * Authored by: Scott Ringwelski <sgringwe@mtu.edu>
  */
 
-public class Noise.RemoveFilesDialog : Gtk.Dialog {
+public class Noise.RemoveFilesDialog : Granite.MessageDialog {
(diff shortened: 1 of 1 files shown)


src/Dialogs/RemoveFilesDialog.vala
```diff
@@ -26,94 +26,61 @@
...
+
+        primary_label.max_width_chars = 65;
+        secondary_label.max_width_chars = 65;
+
+        var to_remove_size = to_remove.size;
+
+        if (to_remove_size > 1) {
+            primary_text = ngettext (
+                "Remove %d Songs From Library?",
+                "Remove %d Songs From Library?",
```
BaptisteGelez: Shouldn't the first one be `Remove One Song From Library?`?

BaptisteGelez: Or at least `Remove %d Song from Library?`

danrabbit: The singular one shouldn't actually get used since it's the totally different string above in this case, but yeah you're right

PR #350, opened 2018-06-18
AlbumListGrid: Create a custom track listbox

danrabbit:
Creates a custom track flowbox for this specific view.

There's a lot of duplicate code in the menu right now, but this is largely because MediaMenu currently requires a GenericList, which this view is not.

screenshot from 2018-09-18 19 13 22 2x


danrabbit:
Yeah this branch is very incomplete. It was mainly just meant for me to figure out the steps I need to take before I can do this. Some things I'm thinking to keep in mind:

It might be worth double checking what interfaces are provided by ContentView and if some should be moved here from GenericList/MusicListView. The MediaMenu should probably call only methods from this interface instead of relying on GenericListMusicListView or implementing the methods itself.

For this particular widget, it might be better to use a FlowBox instead of a list and maybe get more of a responsive layout going on

jeremypw:
Converting to draft as it seems far from ready to merge.

PR #352, opened 2018-06-18
Separate MediaMenu from MusicListView

danrabbit:
This is necessary for eventually using different types of widgets to represent Media such as in #350 

I get this is kind of a big diff. Some of these changes can probably be done in smaller steps if necessary. Let me know :)

### DIFF
--- a/src/Widgets/MediaMenu.vala
+++ b/src/Widgets/MediaMenu.vala
@@ -27,7 +27,8 @@ public class Noise.MediaMenu : Gtk.Menu {
     public bool can_scroll_to_current { get; construct; }
     public ViewWrapper.Hint hint { get; construct ;}
     public GenericList generic_list { get; construct; }
-    public Gtk.MenuItem contractor_entry;
+    public Gtk.MenuItem contractor_entry { get; private set; }
+    public Granite.Widgets.RatingMenuItem rate_media { get; private set; }
 
     private Gtk.MenuItem edit_media;
     private Gtk.MenuItem queue_media;
@@ -47,6 +48,7 @@ public class Noise.MediaMenu : Gtk.Menu {
         edit_media = new Gtk.MenuItem.with_label (_("Edit Song Info…"));
         var file_browse = new Gtk.MenuItem.with_label (_("Show in File Browser…"));
         contractor_entry = new Gtk.MenuItem.with_label (_("Other Actions"));
+        rate_media = new Granite.Widgets.RatingMenuItem ();
         queue_media = new Gtk.MenuItem.with_label (C_("Action item (verb)", "Queue"));
 
         if (can_scroll_to_current) {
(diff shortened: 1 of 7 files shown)


src/Widgets/MediaMenu.vala
```diff
@@ -0,0 +1,327 @@
...
+
+        if (generic_list.parent_wrapper.library.support_playlists () == false) {
+            media_menu_new_playlist.visible = false;
+        }
+        foreach (var playlist in generic_list.parent_wrapper.library.get_playlists ()) {
+            // Don't include this playlist in the list of available options
+            if (playlist == generic_list.playlist)
+                continue;
+
+            if (playlist.read_only == true)
```
BaptisteGelez: Maybe add braces here while we are at it? And this condition and the above may be combined.

src/Widgets/MediaMenu.vala
```diff
@@ -0,0 +1,327 @@
...
+
+    public void popup_media_menu (Gee.Collection<Media> selection) {
+        var media_menu_new_playlist = new Gtk.MenuItem.with_label (_("New Playlist…"));
+        media_menu_new_playlist.activate.connect (media_menu_new_playlist_clicked);
+
+        var add_to_playlist_menu = new Gtk.Menu ();
+        add_to_playlist_menu.append (media_menu_new_playlist);
+
+        if (generic_list.parent_wrapper.library.support_playlists () == false) {
+            media_menu_new_playlist.visible = false;
```
BaptisteGelez: Why not just `media_menu_new_playlist.visible = generic_list.parent_wrapper.library.support_playlists ()`?

src/Widgets/MediaMenu.vala
```diff
@@ -0,0 +1,327 @@
...
+        }
+        add_to_playlist_menu.show_all ();
+        add_to_playlist.submenu = add_to_playlist_menu;
+
+        // if all medias are downloaded already, desensitize.
+        // if half and half, change text to 'Download %external of %total'
+        int temporary_count = 0;
+        int total_count = 0;
+        foreach (var m in selection) {
+            if (m.isTemporary)
```
BaptisteGelez: Braces here too maybe?

src/Widgets/MediaMenu.vala
```diff
@@ -0,0 +1,327 @@
...
+                temporary_count++;
+            total_count++;
+        }
+
+        if (temporary_count < 1) {
+            import_to_library.sensitive = false;
+        } else {
+            import_to_library.sensitive = true;
+            if (temporary_count != total_count)
```
BaptisteGelez: Braces again.

src/Widgets/MediaMenu.vala
```diff
@@ -0,0 +1,327 @@
...
+            } else if (set_rating != m.rating) {
+                set_rating = 0;
+                break;
+            }
+        }
+
+        rate_media.rating_value = set_rating;
+
+        //remove the previous "Other Actions" submenu and create a new one
+        var contractorSubMenu = new Gtk.Menu ();
```
BaptisteGelez: Maybe rename this variable to `contractor_sub_menu`.

danrabbit:
@BaptisteGelez Yep you're right a lot of that is copy/paste. Thanks for being thorough! Should be all fixed :)

PR #385, opened 2018-06-26
Move a few settings to GLib

danrabbit:
All the benefits of #339 without moving the keys around

* cache results when it makes sense
* solves a crapload of terminal warnings


jeremypw:
I have approved the smaller PR.  Settings can be moved later if required.

PR #388, opened 2018-07-03
added documentation target and doc directory

Nine-H:
I think it mostly works, but when you try to build it chokes on some of the comments, not sure what to make of it, maybe I'm doing something wrong.


Nine-H:
whoever thought that it would be a good idea to open documentation comments with /** and format them with whitespace should be shot. I've removed any comments that break valadoc, in the hope that more thoughtful people from the future fix them.

tintou:
looks good

PR #405, opened 2018-08-09
EqualizerPopover: use GLib.Settings

danrabbit:
Use GLib.Settings in EqualizerPopover where convenient.

Several of these settings are still touched in PlayBackManager so they can't be removed until that gets resolved. This is the same reason why I didn't touch the `get_presets ()` method.

Took the opportunity to change a few things to a settings bind

### DIFF
--- a/core/Settings.vala
+++ b/core/Settings.vala
@@ -64,6 +64,7 @@ namespace Noise.Settings {
     }
 
     public class Equalizer : Granite.Services.Settings {
+
         public bool equalizer_enabled { get; set; }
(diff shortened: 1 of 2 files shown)


jeremypw:
I think the binding flag for the sensitive properties should be "GET" not "DEFAULT".

danrabbit:
Oops you're right, it should just be a get for sensitivity.

So what was happening is that the scale area is being set insensitive because the preset is "Automatic" and then because the bind was bidirectional it was turning the EQ back off :package: 

Fixed!

Issue #365, opened 2018-06-22
Make window_state setting a boolean


danrabbit:
There are only two states: maximized or not. So it would be simpler if this was a boolean setting

SneWs:
#410 should fix this.

SneWs:
@jeremypw Should this be closed then?

PR #410, opened 2018-08-18
Make window_state setting a boolean #365

SneWs:
* Fixed - Changed setting name to window-maximized and to be of type boolean.


SneWs:
Roger that, will fix during my evening and update the PR.

SneWs:
Fixed the summary and description as requested, lemme know if further changes are needed.

SneWs:
You're very welcome, and so do I. Really digging what you all are making here.

PR #411, opened 2018-08-19
use system libaccounts-glib.vapi

Nine-H:
we don't need a project specific vapi from juno onwards.


jeremypw:
Is it possible to add a check for a minimum version of vala?

cassidyjames:
This should no longer be blocked, right?

danrabbit:
@Nine-H this seems to be no longer blocked, but it does need a version bump as Cody described above

PR #418, opened 2018-08-30
PlaybackManager: revert queue mechanism

### LINKED ISSUE #347: When restoring the previously played track, medias size is 0
This all occurs in PlaybackManager at about line 250

It looks like when we're restoring the previously played track, we're not restoring a playlist with it which means that the size of the queue playlist is 0 and when we reach the end of the track we have no valid index of what song to play next on repeat. So we divide by zero and crash

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/59788743-when-restoring-the-previously-played-track-medias-size-is-0?utm_campaign=plugin&utm_content=tracker%2F60766398&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60766398&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #347

It's not pretty, but it doesn't try to divide by 0.

Reverts to before [the queue mechanism was rewritten](https://github.com/elementary/music/pull/65), but trying to keep cleanups and such.

### DIFF
--- a/src/LibraryWindow.vala
+++ b/src/LibraryWindow.vala
@@ -775,7 +775,7 @@ public class Noise.LibraryWindow : LibraryWindowInterface, Gtk.Window {
 
         if (p == App.player.queue_playlist) {
             view.set_no_media_alert_message (_("No songs in Queue"), _("To add songs to the queue, use the <b>secondary click</b> on an item and choose <b>Queue</b>. When a song finishes, the queued songs will be played first before the next song in the currently playing list."));
-            App.player.queue_medias (p.medias);
+            App.player.queue_media (p.medias);
         } else if (p == App.player.history_playlist) {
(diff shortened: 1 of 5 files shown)


jeremypw:
Is it not possible to fix divide by zero without reverting a lot of code?  I am not at all familiar with this app so better original author or reviewer review this.

danrabbit:
@jeremypw The problem is that it just punts the issue to another crasher. I tried to fix just the dividing by zero problem, but the review sat there unmerged for months because there were a bunch of other problems introduced. Right now I'm more interested in making sure we don't regress from what was shipped in Loki and it seems like the easiest way to do that is to revert this mechanism so we can come back and improve it with some smaller PRs that are more well reviewed

jeremypw:
Oh well, in that case I agree it should be reverted.

PR #446, opened 2018-10-19
Add orange accent color

cassidyjames:
Switches the accent color to orange, which gives it a nice subtle brand effect imho. Affects overshoot scroll feedback, selected tracks, loading bars, switches, etc.

screenshot from 2018-10-19 16 18 01

screenshot from 2018-10-19 16 18 08


cassidyjames:
lol

PR #452, opened 2018-11-01
ViewSelector: Add shortcut accels to tooltips

danrabbit:


danrabbit:
@cassidyjames my guess is as proper nouns? It looks like Files and the applications menu both use caps for view names. I wouldn't be opposed to changing that though

cassidyjames:
@danrabbit yeah my inclination would be they're not special proper nouns, they're just normal nouns.

PR #463, opened 2018-11-30
Fix loading last playing playlist before search saved state

### LINKED ISSUE #409: Playlist not properly restored: View is empty
If the app is closed while playing a playlist, the when reopened that playlist is selected in the sidebar but the view does not list the songs in that playlist (it is empty).

Mohelm97:
Fixes #409
We should update the search entry from the saved state before loading the current view.


Mohelm97:
@tintou Take a look here mate, I just flip the codes to load the saved search before loading playlist

PR #464, opened 2018-12-04
sort by album artist

### LINKED ISSUE #73: Incorrect track order in album view when a track has multiple artists
When tracks in the same album have different "artist" tags (such as multiple authors for a single song), tracks are ordered first from the "artist" tag and only then from the "track" tag as they should.
As far as i can tell, this only happens in the album view, in the others the correct order is respected.

The artist tag in the problematic files is written like this:
`artist1\\artist2`
which is surely recognized by Music, because in the list view the Artist column divide them with a comma.

Here is an example of expected and actual behaviour:

expected

actual

(note that in Album view you actually view just the title of the song)

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/49555478-incorrect-track-order-in-album-view-when-a-track-has-multiple-artists?utm_campaign=plugin&utm_content=tracker%2F60766398&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60766398&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Nine-H:
Fixes #73

originally thought I'd have to do all sorts of SQL to fix this, turns out no, all this patch does is sort by album_artist when the user tries to sort by artist.


Mohelm97:
@Nine-H Nice catch :D 
But the problem is if we have an album for artist Foo and it contains a song for artist Bar if you order the library by artist Bar will not be at the correct position. and the issue is only present in the album view, I'm not sure but I think there's something orders the songs in the album according to the artist when the app starts.

Nine-H:
@Mohelm97 I want you to check there's an album_artist and artist field in your example album, sorting by artist works for me but I've seen dodgy metadata out there:

master:
screenshot from 2018-12-04 16 12 15

patch:
screenshot from 2018-12-05 01 42 50

Also try clearing your cache `rm ~/.local/share/noise/database_0_3_2.db && rm -rf ~/.cache/noise` 

if you mean you want Bar to show up under B instead of in Foo's album... why? If I want to find all my soulwax tracks then I use a filter because some is just "soulwax", some is "*, and soulwax" and some have soulwax in the title: "* (soulwax remix)"

Mohelm97:
@Nine-H Yeah look at the master order, this is I think the correct order if you order by the artist,
Yes if I have a song for artist Bar but under Foo's album if I sort by the artist I should found him udder B
If you want to fix that soulwax just search for it XD, if the songs under the same album artist sort on it.

Master:
image
Patch:
image

I'm ordering according to the artist alone.

Mohelm97:
@Nine-H Take a look at Slack, we can just set the default sorting for the album view list to track number, and that will fix issue #73 and won't affect any sorting by the song artist.

jeremypw:
@Nine-H Is this PR still alive or has it been abandoned?

jeremypw:
Closing due to inactivity/no response.  Please feel free to reopen.

PR #467, opened 2018-12-05
Sort albums correctly, break up "Various Artist" compilations sorting…

Nine-H:
this is a proof of concept hack, essentially to demonstrate how bad smashing up compilations looks in practice, it's bad code and I put a block on it because

- string literal
- needs localisation to function in LOTE
- don't like the behaviour (needs a switch TBH)

if we go down this path it should probably have more robust code.


Mohelm97:
You can do it the other way around,
If Album artist is prefix of the song artist sort by album artist else by song artist?

Nine-H:
vala strings have a contains method, but I'm not sure how it scales for libraries of 1000s of tracks. I'm okay with whatever method as long as there's a setting to group compilations so they treat the `album` tag like the artist.

Mohelm97:
Yeah Me too, I don't want to add extra stuff in this, I just like it the way it was :(
I think we should leave it to the previous PR and let someone else decide what is better (Just comparing the album artist vs the song artist)

jeremypw:
@Nine-H This appears not have had any development for quite sometime, so closing.  Please feel free to reopen (as a draft) if you wish to continue working on it.

PR #479, opened 2019-02-16
PlaylistsUtils: Use FileChooserNative

### LINKED ISSUE #477: Use FileChooserNative
- [x] https://github.com/elementary/music/blob/bc4bcf18600953b7c3d8b003195cd8645461be6d/src/Dialogs/FileNotFoundDialog.vala#L99
- [x] https://github.com/elementary/music/blob/5d7008fa2951a77efcd0f266c9e7d50769ad222e/core/Utils/PlaylistsUtils.vala#L273
- [x] https://github.com/elementary/music/blob/5d7008fa2951a77efcd0f266c9e7d50769ad222e/core/Utils/PlaylistsUtils.vala#L367
- [x] https://github.com/elementary/music/blob/5d7008fa2951a77efcd0f266c9e7d50769ad222e/src/Dialogs/SetMusicFolderConfirmation.vala#L113
- [x] https://github.com/elementary/music/blob/c670ab5ea7d009d73a17d34fa35dd903a7c04b21/src/Views/Wrappers/MusicViewWrapper.vala#L143
- [x] https://github.com/elementary/music/blob/52af19c1d7443235eb4abcef45fbee1c2d05095a/src/Widgets/AlbumListGrid.vala#L247
- [x] https://github.com/elementary/music/blob/b30c307755613189ba7a5cf51ef3cfb4334c5a4f/src/LibraryWindow.vala#L981

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/70105602-use-filechoosernative?utm_campaign=plugin&utm_content=tracker%2F60766398&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60766398&utm_medium=issues&utm_source=github).
</bountysource-plugin>

danrabbit:
Fixes #477


Mohelm97:
@danrabbit For some reason when we use FileChooserNative the signals doesn't work at all, so changing the file extension if the user changed the filter will not work :(

danrabbit:
@Mohelm97 oh yup you're right. According to the docs, we can't modify the chooser once it has been opened, so this feature has to go.

Mohelm97:
@danrabbit I've tried today to implement it in the chooser but we can't :( as you can't get the pattern from Gtk.FileFilter, I'm not sure if we should just lose this feature or stay in the old way of using filechooserdialog?

danrabbit:
@Mohelm97 Yes, if this is a desirable feature we should implement it in the chooser itself. It shouldn't be in the app anyways. So it should be removed here and added to the chooser

Issue #491, opened 2019-03-08
Removing from queue isn't working


Mohelm97:
To reproduce:
1. Queue a list
2. Remove the next song
3. Click next (the removed song will be played!)

PR #492, opened 2019-03-08
Fix removing from queue

Mohelm97:
Fix: #491

### DIFF
--- a/src/PlaybackManager.vala
+++ b/src/PlaybackManager.vala
@@ -133,6 +133,38 @@ public class Noise.PlaybackManager : Object {
 
     public void unqueue_media (Gee.Collection<Media> to_unqueue) {
         queue_playlist.remove_medias (to_unqueue);
+
+        if (_current.size - to_unqueue.size > 0){
+            bool currnet_media_changed = false;
+            while (current_media != null && to_unqueue.contains (current_media)) {
+                currnet_media_changed = true;
+                current_index++;
+                if (current_index >= _current.size) {
+                    current_index = 0;
+                }
+                if (_current_shuffled.size == 0) {
+                    current_media = _current.get (current_index);
+                } else {
+                    current_media = _current_shuffled.get (current_index);
+                }
+            }
+
+            if (currnet_media_changed && current_media != null) {
+                play_media (current_media);
+            }
+
+        } else {
+            current_media = null;
+            stop_playback ();
+        }
+
+        _current.clear();
+        foreach (var q in queue_playlist.medias) {
+            debug ("REQUEUED: %s", q.title);
+            add_to_current (q);
+        }
+
+        reshuffle ();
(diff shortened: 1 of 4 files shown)


src/PlaybackManager.vala
```diff
@@ -133,6 +133,38 @@ public class Noise.PlaybackManager : Object {
 
     public void unqueue_media (Gee.Collection<Media> to_unqueue) {
         queue_playlist.remove_medias (to_unqueue);
+
+        if (_current.size - to_unqueue.size > 0){
+            bool currnet_media_changed = false;
```
danrabbit: Probably wanna fix this typo ;)
```suggestion
            bool current_media_changed = false;
```

src/Views/Wrappers/ViewWrapper.vala
```diff
@@ -371,7 +371,6 @@ public abstract class Noise.ViewWrapper : Gtk.Grid {
             set_active_view (ViewType.GRID);
         else if (has_list_view) {
             view_stack.visible_child = list_view;
-            list_view.list_view.scroll_to_current_media (true);
```
danrabbit: Is there a reason this is removed? I think we did this so that we always keep the currently playing song in view

Mohelm97: I've removed this because if you delete a song from the queue it will jump to the currently played song which I don't think is the correct behavior if i want to edit the queue (will you can try both to see which is better :D)

Mohelm97:
@jeremypw Are you sure because this is the behaviour of this branch
Screen record from 2019-04-26 01 06 27
And this is the master
Screen record from 2019-04-26 01 09 43

jeremypw:
@Mohelm97 OK, I'll check again.

jeremypw:
I cannot tell from your examples what song is heard.  The way I tested the branch was this:

* Add three songs to the queue.
* Start playing the first song
* Remove second song from queue
* Press `>>`
* The removed song plays

This happens regardless of whether shuffle mode is on.

jeremypw:
OK, I have located the problem - your code is OK but is not getting called by the context menu item "Remove from queue" (!).

You need 
```
            case ViewWrapper.Hint.QUEUE:
                generic_list.playlist.remove_medias (selected_media);
                App.player.unqueue_media (selected_media);
                break;
```

in `remove_media_clicked ()` in  `MediaMenu.vala`

Mohelm97:
@jeremypw Ops, sorry I didn't check the context menu, shame on me :$ I will fix it (y)

jeremypw:
This fixes the linked issue.  There is still a related bug regarded removing items from a custom playlist - the removed item still plays when the "next" button is pressed.  However that can be fixed in another PR, hopefully not repeated too much code.

PR #509, opened 2019-06-26
Spelling: An Error Occurred → Error

comradekingu:


danrabbit:
I don't think we want to change a full sentence to just a single word. For dialog primary text, we still want  a complete sentence

Issue #334, opened 2018-05-29
Music isn't displaying icon in the AppCenter


btd1337:
version: Juno Beta

![captura de tela de 2018-05-29 01-56-25](https://user-images.githubusercontent.com/11943860/40638825-f3ad4840-62e3-11e8-9bee-96e1d7f46ba6.png)

ryonakano:
I can reproduce this

![screenshot from 2018-05-29 19-28-12](https://user-images.githubusercontent.com/26003928/40653676-d4eedfe8-6376-11e8-9d0c-5cfeee9d07db.png)

liamkinne:
Also having this issue. I'll have a look

4jNsY6fCVqZv:
@danrabbit @cassidyjames The issue is still relevant.Any idea how to solve it?

ryonakano:
@4jNsY6fCVqZv The PR #511 should fix this issue, I've never tested it though.

4jNsY6fCVqZv:
@ryonakano It would be nice if the PR could fix the issue. Then all we need is a solution for the other apps mentioned.

ryonakano:
Closing since the issue seems to be gone on Hera

PR #511, opened 2019-07-13
Fix data files so Music has an icon in AppCenter

lainsce:
Should merge #512 soon after this one.

Overall, this PR will fix most problems in the data folder.

- Fixed some lack of linking the data folder in `meson.build` files, and also the `gresource` file,
so that Music has an icon in AppCenter.
- Also removed the duplicate icons in the data folder as they were unused.
- Fix tabbing in the appdata.xml file as XML is tab-sensitive.

Despite the Travis CI :x: this is stable.


4jNsY6fCVqZv:
Thanks for your PR!
Do you think you can fix that for the other apps as well? https://github.com/elementary/music/issues/334#issuecomment-516761373

lainsce:
I'll gradually do it.

4jNsY6fCVqZv:
Thanks for working on that! Is the procedure for the other mentioned apps similar to your fix for music?

lainsce:
It should be.

4jNsY6fCVqZv:
But not for Firefox and GStreamer Multimedia Codecs as third party packages, right?

Hmm, that would be my first fix ever. At the moment, for example, the file structure of the projects also looks different. If you guide me, I can try to prepare PR's for Screenshot and Code.

When will your PR be merged?

lainsce:
> But not for Firefox and GStreamer Multimedia Codecs as third party packages, right?

Yeah, those are outside my reach as an elementary developer.

> When will your PR be merged?

Whenever there is an approving review.

davidmhewitt:
There are a number of seemingly unrelated changes here, this is very difficult to review. 

There's probably some good stuff in here in terms of tidying stuff up but it would be better if you separated this out into different pull requests. It's hard to review changes to translation files, deletion of icons, massive whitespace and markup changes in XML (which by the way isn't indentation sensitive), and inexplicably no longer installing the icons to the `hicolor/apps` folder anymore all in one PR?

Given that the issue with icons not appearing in AppCenter is caused by the `appstream-data-pantheon` package (see https://github.com/elementary/appstream-data/issues/13), and I've tested and proven a fix where the icons for Music appear in AppCenter without this being merged, I'd be tempted to close this so it can be broken down into individual cleanup PRs.

danrabbit:
Yeah I agree with @davidmhewitt. This is (almost certainly) not the cause of those issues and it would be better to provide smaller, individual PRs for other cleanup

PR #517, opened 2019-09-04
Remove the iPod plugin

danrabbit:
LibiMobileDevice hasn't had a release in like 5 years


tintou:
It is still actively developed though :/

danrabbit:
You're right, it looks like the Ubuntu package uses git snapshots and not just releases

PR #523, opened 2019-09-04
LibrariesManager: Remove unused signals, methods, reduce variable scope

danrabbit:


core/LibrariesManager.vala
```diff
@@ -31,22 +30,17 @@ public class Music.LibrariesManager : GLib.Object {
      * Headless playlists are playlists that are not linked to a library.
      */
     public signal void add_headless_playlist (StaticPlaylist playlist);
-
-    public signal void library_removed (Library library);
```
jeremypw: Seems odd that this signal (and the corresponding function) is not needed for a libraries manager.  Pluggable devices can add a library but do not seem to remove it when unplugged??

jeremypw:
I think you need to justify these changes to the core interface.  Is there an identified issue in exposing these methods/signals or can you be sure they will never be useful for future plugins?

PR #524, opened 2019-09-04
PluginInterface: Remove unused and cleaup

danrabbit:


core/PluginInterface.vala
```diff
@@ -31,14 +30,11 @@
  */
 
 public class Music.Plugins.Interface : Object {
-    Manager manager;
+    public Manager manager { get; construct; }
```
jeremypw: I think this back reference can be unowned.

jeremypw:
Are there any 3rd party plugins we need worry about?

danrabbit:
@jeremypw I don't think there exist any 3rd party plugins, no

PR #526, opened 2019-09-04
ViewInterface: Remove unused get_hint

danrabbit:


ryonakano:
I guess we can also remove the same method in the following line:

https://github.com/elementary/music/blob/4d4b81510ade9c485a65a72bd5e015101370c2bf/src/Views/ListView/ListView.vala#L215-L218

PR #559, opened 2019-12-05
Fix lint errors

jeremypw:


cassidyjames:
Lots of warnings still, but all errors have been taken care of.

PR #562, opened 2020-01-03
Properly disable equalizer levels

markodp:
If you had equalizer turned off and selected anything except "Automatic", level sliders were still enabled.


markodp:
Do you mean in /data/io.elementary.music.appdata.xml.in? Under version 5.0.5?

PR #571, opened 2020-02-22
Enable playback buttons on startup if media is available

vjr:
Fixes #570

PR #573, opened 2020-02-27
Fix error on clicking previous button for the first time

vjr:
Fix an error similar to https://github.com/elementary/music/pull/571#pullrequestreview-363042683 when clicking the previous button for the first time, before clicking play.


vjr:
@jeremypw can you review and merge?

PR #578, opened 2020-03-18
Use GLib.Settings for Shuffle and Repeat

danrabbit:


src/PlaybackManager.vala
```diff
@@ -175,15 +188,15 @@ public class Music.PlaybackManager : Object {
...
+        if (Music.App.settings.get_enum ("shuffle-mode") == ShuffleMode.OFF) {
             return _current.get (index_in_current);
         } else {
             return _current_shuffled.get (index_in_current);
         }
     }
 
     public Gee.Collection<Media> get_current_media_list () {
-        if (Settings.Main.get_default ().shuffle_mode == Music.Settings.Shuffle.OFF) {
+        if (Music.App.settings.get_enum ("shuffle-mode") == ShuffleMode.OFF) {
```
jeremypw: This looks odd - although it is a correct modification of the original.  Why is the shuffled collection returned when ShuffleMode is OFF?   As far as I can see, this function is currently only called in one place and then only to determine the size of the collection so there may not be any consequence if this is indeed an error.

PR #603, opened 2020-09-23
StatusBar: use flat style

danrabbit:
While we're here, code style and make sure primary action has a label


cassidyjames:
With this change, the EQ tooltip now shows for the "Add Playlist" button, too; I'm guessing it's being set on the entire toolbar, and before it wasn't as noticable because the "Add Playlist" button had its own tooltip.

cassidyjames (approved):

PR #604, opened 2020-09-25
Less intrusive import error handling

### LINKED ISSUE #470: Don't suggest to move unrecognized files to Trash on each run
Part of my music is stored in a format that this application does not recognize (`.it` - Impulse Tracker files). Every time I open Music, it suggests to move these "corrupted files" to Trash:

io elementary music_276

I would suggest to only see files as "corrupted" when they have a recognized file extension, but still fail to load, since when all unrecognized files are marked as such this helpful message is actually quite dangerous.

In addition, it would be nice if the feature could be turned off entirely. I generally would expect my music player to just ignore files it doesn't recognize.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/67745140-don-t-suggest-to-move-unrecognized-files-to-trash-on-each-run?utm_campaign=plugin&utm_content=tracker%2F60766398&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F60766398&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jeremypw:
Fixes #470 

Screenshot from 2020-09-28 12 29 20@2x

 - [x] Use toast for initial warning with button to access more information
 - [x] Show summary of different types of errors found in dialog
 - [x] Button to open filemanager at library location
 - [X] Provide setting to turn off warning
 
Ideally there should be an option to install missing plugin(s) but that will be left for a future PR.

The styling is pretty basic but can be improved if this is accepted.


danrabbit:
Hey thanks for working on this and trying to provide better error feedback here! I agree that offering to trash these files is probably not great and showing them in Files is probably more productive.

I'm not sure a toast is a good option for this since toasts automatically expire and an import may be a long-running operation where the user steps away from the computer. A dialog is probably still appropriate for errors, especially if there is an option to always ignore errors

There should probably be some secondary text that explains the implications of an import error. For example that these files won't appear or be playable in Music.

I'm not sure always showing the library is the most helpful here, especially if the import is being performed from a different folder. This button should probably say something like "Show Files" and open the import location. It should probably be in the primary action position in the dialog and not floating in the center of the dialog.

"Ignore" would probably be a better dismiss action since it makes it explicit that you're electing to take no action on this problem and not just that you're closing the window.

Regarding the checkbox, I wonder if we really need this feature. It sounds like the original issue is that the import error dialog is shown while performing an automatic import. We probably don't want this dialog to not appear when performing a manual import. Failing silently is probably not expected for explicit action. But in the case of an automatic library rescan, failing silently is probably not a big deal. So it might solve the problem to fail silently on automatic imports without adding this new setting

src/LibraryWindow.vala
```diff
@@ -1260,4 +1279,106 @@ public class Music.LibraryWindow : LibraryWindowInterface, Hdy.ApplicationWindow
...
+
+                string secondary_text = "";
+                string invalid_text = "";
+                string error_text = "";
+                string timeout_text = "";
+                string busy_text = "";
+                string missing_plugin_text = "";
+
+                if (invalid_files.size > 0) {
+                    invalid_text = _("Invalid uri: %i").printf (invalid_files.size) + "\n";
```
danrabbit, line 1312: It's probably not clear to most people what a "URI" means

src/LibraryWindow.vala
```diff
@@ -1260,4 +1279,106 @@ public class Music.LibraryWindow : LibraryWindowInterface, Hdy.ApplicationWindow
...
+                if (error_files.size > 0) {
+                    error_text = _("Could not be read and may be damaged: %i").printf (error_files.size) + "\n";
+                }
+
+                if (timeout_files.size > 0) {
+                    timeout_text = _("Took too long to load: %i").printf (timeout_files.size) + "\n";
+                }
+
+                if (busy_files.size > 0) {
+                    busy_text = _("Already being used: %i").printf (busy_files.size) + "\n";
```
danrabbit, line 1324: I'm not sure it's clear what this means. Already being used by who or what?

src/LibraryWindow.vala
```diff
@@ -1260,4 +1279,106 @@ public class Music.LibraryWindow : LibraryWindowInterface, Hdy.ApplicationWindow
...
+                }
+
+                if (missing_plugin_files.size > 0) {
+                    missing_plugin_text = _("Required gstreamer plugin not installed: %i").printf (missing_plugin_files.size) + "\n";
```
danrabbit, line 1328: Gstreamer is a niche technical detail that probably doesn't mean anything to most people. Mentioning "codecs" might be a little more generally understandable, but maybe there's a better word for this too. Maybe something in the direction of "Additional software codecs are needed:"

danrabbit (changes requested):
Looking at some of the error messaging, it's not super clear what these things mean or what can be done about it and just listing the number of files doesn't really explain which files are causing issues. An error dialog should ideally help the user to address the problem.

For example if there is an "invalid uri" what does that mean to someone who is trying to listen to their music and how can they fix it?

If something "took too long to load" should they wait and try again? Could this issue be resolved with a better network connection? Why do we have a time limit on how long it takes to load something, shouldn't we just wait for it to load?

jeremypw:
Thanks for a lot of useful points - I'll get to work on them.  With regard to replacing the toast, do you suggest launching the dialog directly or perhaps putting some kind of error warning icon into the headerbar or somewhere which launches the dialog when pressed (and then disappears)?

Where a file is reported as busy (probably very rare in practice) I am not sure how to find out why - unless it is in the original error message. I used my own (draft) error summaries instead of outputting each error message as received.  Is it worth providing access to the full error details?

danrabbit:
Closing since this targets old Music

PR #608, opened 2020-10-09
Music: Give Top Display more contrast.

BAProductions:
Screenshot from 2020-10-09 13 34 16@2x
Screenshot from 2020-10-09 13-53-02@2x

By switching from
Gtk.STYLE_CLASS_TITLEBAR to Gtk.STYLE_CLASS_ENTRY  the application tends look nicer, But with third party themes the look bake.

### DIFF
--- a/src/Widgets/TopDisplay.vala
+++ b/src/Widgets/TopDisplay.vala
@@ -75,7 +75,7 @@ public class Music.TopDisplay : Gtk.Stack {
         add_named (action_grid, "action");
         add_named (time_grid, "time");
         add_named (empty_grid, "empty");
-        get_style_context ().add_class (Gtk.STYLE_CLASS_TITLE);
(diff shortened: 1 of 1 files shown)


danrabbit (changes requested):
So, this method won't actually work anymore with the stylesheet in elementary OS 6. You'll need to set the css_name property in a `class construct` block. Then it will look like this:

Screenshot from 2020-10-13 12 33 30

I'm not really a fan of how much larger this makes the headerbar. It looks like in this style contrast is raised to 12.6:1 from 10.5:1, but since WCAG AAA is 7:1 it doesn't really feel like this is actually necessary

danrabbit:
I think I'm gonna close this as "Design conflict" considering contrast is already quite high and this has that trade off of really increasing the headerbar size

PR #618, opened 2021-02-02
Replace Gtk.Dialog with Granite.Dialog

danrabbit:


cassidyjames:
Looks like dialogs still need action area spacing cleanup, but this does what it says.

cassidyjames (approved):

PR #624, opened 2021-02-10
SetMusicFolderConfirmation: Fix action area margin, use response

danrabbit:
* Bump copyright
* Fix margins around the action area
* Use response connect to save a few lines


jeremypw (approved):
No regressions

PR #628, opened 2021-04-05
RDNN Icons and MPRIS

Marukesu:
Update icons and use rdnn since multimedia-audio-player is not a freedesktop defined icon, also rdnn mpris so we don't need the `--own-name` on the flatpak manifest


danrabbit:
Can you use`git mv` here so the diff doesn't look so massive?

Marukesu:
I did use git mv, but seems they ignored it.

danrabbit (approved):

PR #631, opened 2021-04-23
FileNotFoundDialog: Fix a typo

ryonakano:
I guess this sentence should end with a period, not a question mark.


jeremypw (approved):
I agree.

PR #633, opened 2021-05-02
Add launchable tag to appdata

City-busz:
This is needed for appstream-generator to find the associated desktop file.

See also:
- https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-launchable
- https://github.com/ximion/appstream-generator/issues/88


jeremypw (approved):
Makes sense. Thanks for the contribution!

PR #636, opened 2021-05-20
Fix TileRenderer image scaling when using HiDPI

EFDos:


danrabbit:
Thanks for your branch! We're doing a from-scratch rewrite of Music and this no longer applies to the new code base. I'm closing this PR since the issue can't be reproduced in the new version of Music.

PR #642, opened 2021-07-15
desktop: No longer mention CD

ryonakano:
Just a tiny proposal, but I think we may want to remove the word "CD" from search words since Music no longer supports CDs: #577


danrabbit (approved):
Yup good call!

PR #643, opened 2021-07-15
Remove LastFM plugin

danrabbit:
This relies on online accounts stuff that doesn't exist anymore in Odin


cassidyjames (changes requested):
Don't forget translations https://github.com/elementary/music/blob/ccf9f28975ad9561c2461e61028c46b73bcd51f3/po/POTFILES#L39-L43

cassidyjames (approved):

PR #652, opened 2021-08-17
Queuelist

danrabbit:


src/MainWindow.vala
```diff
@@ -23,10 +42,40 @@ public class Music.MainWindow : Hdy.ApplicationWindow {
...
 
-        var grid = new Gtk.Grid ();
-        grid.attach (headerbar, 0, 0);
-        grid.attach (now_playing_view, 0, 1);
+        var now_playing = new Gtk.Grid ();
+        now_playing.attach (headerbar, 0, 0);
+        now_playing.attach (now_playing_view, 0, 1);
+
+        var paned = new Gtk.Paned (Gtk.Orientation.HORIZONTAL) {
+            position = 350
```
danrabbit, line 50: Need to save this state

src/MainWindow.vala
```diff
@@ -7,6 +7,25 @@ public class Music.MainWindow : Hdy.ApplicationWindow {
...
+        queue_header_context.add_class (Granite.STYLE_CLASS_DEFAULT_DECORATION);
+        queue_header_context.add_class (Gtk.STYLE_CLASS_FLAT);
+
+        var queue_listbox = new Gtk.ListBox () {
+            expand = true,
+            sensitive = false
+        };
+        queue_listbox.bind_model (PlaybackManager.get_default ().queue_liststore, create_queue_row);
+
+        var queue = new Gtk.Grid ();
```
danrabbit, line 25: This should probably have view class. Looks like headerbar inside view or even headerbar.view doesn't actually work

src/MainWindow.vala
danrabbit: Looks like I'm not saving any state yet so nvm

danrabbit:
Better than no queue :shrug:

PR #657, opened 2021-08-19
Create AudioObject

danrabbit:
Couple of motivations here:

* Makes the relationship between the NowPlaying view and PlaybackManager a lot simpler
* Gives us a way to store metadata when we find it and communicate that to the queue


src/Views/NowPlayingView.vala
```diff
@@ -61,21 +61,29 @@ public class Music.NowPlayingView : Gtk.Grid {
...
+            if (playback_manager.current_audio != null) {
+                playback_manager.current_audio.bind_property ("artist", artist_label, "label");
+                playback_manager.current_audio.bind_property ("title", title_label, "label");
+
+                playback_manager.current_audio.notify["pixbuf"].connect (() => {
+                    var pixbuf = playback_manager.current_audio.pixbuf;
+                    var scaled = pixbuf.scale_simple (200, 200, Gdk.InterpType.BILINEAR);
+
+                    album_image.image.pixbuf = scaled;
+                });
```
danrabbit, line 75: Probably need to make sure to unbind/disconnect if it changes?

PR #658, opened 2021-08-19
PlaybackManager: Notify when tracks are queued

danrabbit:
Nice bit of feedback when we queue tracks from another app

Give the queue notifications a unique ID so that they replace each other


jeremypw (approved):
Works as expected e.g Opening music files from Files while Music is already playing in a different workspace.

Refinements like being able to see which files were added can be done a further PR if required.

PR #668, opened 2021-11-24
Use FileChooserNative to choose music folder

jeremypw:
This changes the one remaining place that does not use the filechooser portal to use it.  Also changes the environment so that the portal is in fact used even if not Flatpak'd, in line with other elementary core apps such as Code.

There seems to be no way to get FileChooserButton to use the FileChooserNative interface and moreover this widget does not seem to appear in Gtk4 so it is replaced with an ordinary button.

Needs input from UX team as to what the desired appearance is.

MusicFolderLocationButton

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -52,6 +52,8 @@ public class Music.App : Gtk.Application {
         GLib.Intl.bind_textdomain_codeset (GETTEXT_PACKAGE, "UTF-8");
         GLib.Intl.textdomain (GETTEXT_PACKAGE);
 
+        Environment.set_variable ("GTK_USE_PORTAL", "1", true);
+
         weak Gtk.IconTheme default_theme = Gtk.IconTheme.get_default ();
         default_theme.add_resource_path ("/io/elementary/music");
 


tintou:
This can't work in a sandboxed app though, we can't have access to this folder after relaunch

jeremypw:
@tintou Do you mean the music folder?  Surely we can get it from the settings?

jeremypw:
I assume the music folder will be in persistent accessible storage of course.

tintou:
Music don't have access to all the folders on the system so you can have the path to the folder after using the portal but the path won't be remounted into the sandbox on launch

jeremypw:
At the moment, Music has full access to `/home` and read only access to the rest of the system.  I am building a Flatpak version to see what the limitations are. I assume Music will eventually be supplied as a flatpak? I do not see that changing the FileChooser to FileChooserNative will influence the issue you mention will it?  Either way its still in a sandbox (?)

tintou:
I'm more questioning the setting as a whole, not this change in particular

jeremypw:
OK, I did not have any problem using and switching between two music folders located below `\home` when running as a Flatpak.

I could also import music from a separate partition into the music folder.

So is the problem that the user could choose a music folder that is read-only or only exists in the sandbox?

tintou:
The issue is that Music shouldn't be able to see the full host, even read-only.

Marukesu:
> There seems to be no way to get FileChooserButton to use the FileChooserNative interface

Gtk.FileChooserButton uses the Gtk.FileChooserNative by default, maybe it just don't listen to the GTK_USE_PORTAL env?

> The issue is that Music shouldn't be able to see the full host, even read-only.

I would prefer to have music only permanently access xdg-music, i believe videos already only has permanent access to xdg-videos.

Marukesu:
> This can't work in a sandboxed app though, we can't have access to this folder after relaunch

Are your sure about that? in the begin of the year i tried a "no --filesystem" build of code and it still have access to exported folders and files after a relaunch.

jhaygood86:
@Marukesu That's what I would expect. The FileChooser portal registers the selected URI with the Documents portal with the persistent flag set to true which **should** make it survive app re-launches.

BrainBlasted:
That is the case, yes.

jeremypw:
> Gtk.FileChooserButton uses the Gtk.FileChooserNative by default

I was going by what https://valadoc.org/gtk+-3.0/Gtk.FileChooserButton.html seems to say.  Also I couldn't see that widget in Gtk4 (on Valadoc) although it is not deprecated.  Will it still be available in Gtk4.  Even if Music remains non-Flatpak'd for now, using the elementary portal will improve consistency with other apps.

jeremypw:
I agree that Flatpaks should have the minimum file access to fulfil their function but I prefer to keep my music files on a separate internal disc or partition so presumably that would mean changing my xdg-music setting?  That would not be so easy for a less experienced user.

jeremypw:
Looks like @danrabbit is porting Music to Gtk4 atm anyway ...

danrabbit:
Closing since this targets old Music

PR #669, opened 2021-11-25
Port to Gtk4

danrabbit:
Todo:
- [x] Extra titles in headerbar
- [x] Fix commented code
- [x] https://github.com/elementary/flatpak-platform/pull/83

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -31,22 +31,22 @@ public class Music.Application : Gtk.Application {
             default_width = 650,
             title = _("Music")
         };
-        main_window.show_all ();
+        main_window.present ();
 
-        Gtk.IconTheme.get_default ().add_resource_path ("/io/elementary/music");
+        Gtk.IconTheme.get_for_display (Gdk.Display.get_default ()).add_resource_path ("/io/elementary/music");
 
-        var granite_settings = Granite.Settings.get_default ();
-        var gtk_settings = Gtk.Settings.get_default ();
+        // var granite_settings = Granite.Settings.get_default ();
+        // var gtk_settings = Gtk.Settings.get_default ();
 
-        gtk_settings.gtk_application_prefer_dark_theme = (
-            granite_settings.prefers_color_scheme == Granite.Settings.ColorScheme.DARK
-        );
+        // gtk_settings.gtk_application_prefer_dark_theme = (
+        //     granite_settings.prefers_color_scheme == Granite.Settings.ColorScheme.DARK
+        // );
 
-        granite_settings.notify["prefers-color-scheme"].connect (() => {
-            gtk_settings.gtk_application_prefer_dark_theme = (
-                granite_settings.prefers_color_scheme == Granite.Settings.ColorScheme.DARK
-            );
-        });
+        // granite_settings.notify["prefers-color-scheme"].connect (() => {
+        //     gtk_settings.gtk_application_prefer_dark_theme = (
+        //         granite_settings.prefers_color_scheme == Granite.Settings.ColorScheme.DARK
+        //     );
+        // });
     }
 
     protected override void open (File[] files, string hint) {
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -35,18 +35,18 @@ public class Music.Application : Gtk.Application {
 
         Gtk.IconTheme.get_for_display (Gdk.Display.get_default ()).add_resource_path ("/io/elementary/music");
 
-        // var granite_settings = Granite.Settings.get_default ();
-        // var gtk_settings = Gtk.Settings.get_default ();
+        var granite_settings = Granite.Settings.get_default ();
+        var gtk_settings = Gtk.Settings.get_default ();
 
-        // gtk_settings.gtk_application_prefer_dark_theme = (
-        //     granite_settings.prefers_color_scheme == Granite.Settings.ColorScheme.DARK
-        // );
+        gtk_settings.gtk_application_prefer_dark_theme = (
+            granite_settings.prefers_color_scheme == Granite.Settings.ColorScheme.DARK
+        );
 
-        // granite_settings.notify["prefers-color-scheme"].connect (() => {
-        //     gtk_settings.gtk_application_prefer_dark_theme = (
-        //         granite_settings.prefers_color_scheme == Granite.Settings.ColorScheme.DARK
-        //     );
-        // });
+        granite_settings.notify["prefers-color-scheme"].connect (() => {
+            gtk_settings.gtk_application_prefer_dark_theme = (
+                granite_settings.prefers_color_scheme == Granite.Settings.ColorScheme.DARK
+            );
+        });
     }
 
     protected override void open (File[] files, string hint) {
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -39,9 +39,8 @@ public class Music.MainWindow : Gtk.ApplicationWindow {
             child = queue_listbox
         };
 
-        var queue = new Gtk.Grid () {
-            css_classes = {"view"}
-        };
+        var queue = new Gtk.Grid ();
+        queue.add_css_class ("view");
         queue.attach (queue_header, 0, 0);
         queue.attach (scrolled, 0, 1);
 
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -65,13 +65,14 @@ public class Music.MainWindow : Gtk.ApplicationWindow {
         now_playing.attach (headerbar, 0, 0);
         now_playing.attach (now_playing_view, 0, 1);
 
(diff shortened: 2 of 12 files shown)


src/Application.vala
```diff
@@ -31,22 +31,22 @@ public class Music.Application : Gtk.Application {
             default_width = 650,
             title = _("Music")
         };
-        main_window.show_all ();
+        main_window.present ();
 
-        Gtk.IconTheme.get_default ().add_resource_path ("/io/elementary/music");
+        Gtk.IconTheme.get_for_display (Gdk.Display.get_default ()).add_resource_path ("/io/elementary/music");
```
BrainBlasted, line 36: Alternatively, setting `resource-base-path` on the app before startup should do the same thing for you

danrabbit: Thanks for the tip!!

src/MainWindow.vala
```diff
@@ -66,20 +62,19 @@ public class Music.MainWindow : Hdy.ApplicationWindow {
         now_playing.attach (headerbar, 0, 0);
         now_playing.attach (now_playing_view, 0, 1);
 
-        var now_playing_handle = new Hdy.WindowHandle ();
-        now_playing_handle.add (now_playing);
+        // var now_playing_handle = new Hdy.WindowHandle ();
```
BrainBlasted, line 65: We have GtkWindowHandle now

src/MainWindow.vala
```diff
@@ -3,61 +3,57 @@
  * SPDX-FileCopyrightText: 2021 elementary, Inc. (https://elementary.io)
  */
 
-public class Music.MainWindow : Hdy.ApplicationWindow {
+public class Music.MainWindow : Gtk.ApplicationWindow {
```
BrainBlasted, line 6: AdwApplicationWindow? Or, set the titlebar to a blank widget. Otherwise we have this: https://imgur.com/zsvezd0

src/MainWindow.vala
```diff
@@ -3,61 +3,57 @@
...
+        // }
 
-            Gtk.StyleContext.add_provider_for_screen (Gdk.Screen.get_default (), css_provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
-        } catch (Error e) {
-            critical ("Can't set accent color: %s", e.message);
-        }
-
-        var queue_header = new Hdy.HeaderBar () {
+        var queue_header = new Adw.HeaderBar () {
+            css_classes = {"default-decoration", "flat"},
```
Exalm, line 18: css_classes is generally a bad idea - it replaces css classes isntead of adding them. It's not too bad _here_, but e.g. will break buttons - as buttons have `.image-button` and other classes.

```vala
queue_header.get_css_class ("default-decoration");
queue_header.get_css_class ("flat");
```

src/MainWindow.vala
```diff
@@ -3,61 +3,57 @@
...
+            vexpand = true
         };
         queue_listbox.bind_model (PlaybackManager.get_default ().queue_liststore, create_queue_row);
-        queue_listbox.set_placeholder (queue_placeholder);
+        // queue_listbox.set_placeholder (queue_placeholder);
 
-        var scrolled = new Gtk.ScrolledWindow (null, null);
-        scrolled.add (queue_listbox);
+        var scrolled = new Gtk.ScrolledWindow () {
+            child = queue_listbox
```
Exalm, line 37: One annoying thing: the implicit GtkViewport gtk inserts here does _not_ have `scroll-to-focus=true` and you need to add one manually.

src/MainWindow.vala
Exalm: You probably don't need a window handle here at all. GtkHeaderBar and AdwHeaderBar are both draggable automatically.

src/Views/NowPlayingView.vala
```diff
@@ -25,17 +26,16 @@ public class Music.NowPlayingView : Gtk.Grid {
...
+        var play_pause_image = new Gtk.Image.from_icon_name ("media-playback-start-symbolic") {
+            pixel_size = 24
+        };
 
         var play_button = new Gtk.Button () {
             action_name = Application.ACTION_PREFIX + Application.ACTION_PLAY_PAUSE,
+            css_classes = {"flat"},
             halign = Gtk.Align.CENTER,
-            image = play_pause_image
+            child = play_pause_image
```
Exalm, line 37: GtkButton has `icon-name` now. :) But here: `new Gtk.Button.from_icon_name`. And this will also add `.image-button` style class, you will have to remove this `css_classes` as it breaks that.

src/Views/NowPlayingView.vala
```diff
@@ -71,7 +71,7 @@ public class Music.NowPlayingView : Gtk.Grid {
                     var pixbuf = playback_manager.current_audio.pixbuf;
                     var scaled = pixbuf.scale_simple (200, 200, Gdk.InterpType.BILINEAR);
 
-                    album_image.image.pixbuf = scaled;
+                    album_image.image.set_from_pixbuf (scaled);
```
Exalm, line 74: Probably want to replace pixbufs with textures later.

src/Widgets/AlbumImage.vala
```diff
@@ -28,10 +27,11 @@ public class Music.AlbumImage : Gtk.Grid {
             pixel_size = 200
         };
 
+        css_classes = {"card"};
         halign = Gtk.Align.CENTER;
         valign = Gtk.Align.CENTER;
         height_request = 200;
         width_request = 200;
-        add (image);
+        attach (image, 0, 0);
```
Exalm, line 35: `Gtk.Box`, `append()`? Unless you actually have multiple columns and rows here - but I've seen elementary apps using a lot of unnecessary grids so saying just in case.

src/Widgets/SeekBar.vala
```diff
@@ -61,35 +61,35 @@ public class Music.SeekBar : Gtk.Grid {
...
         scale = new Gtk.Scale.with_range (Gtk.Orientation.HORIZONTAL, 0, 1, 0.1) {
+            css_classes = {"accent"},
             draw_value = false,
             hexpand = true
         };
-        scale.get_style_context ().add_class (Granite.STYLE_CLASS_ACCENT);
 
-        scale.button_press_event.connect (() => {
-            scale.value_changed.connect (scale_value_changed);
+        // scale.button_press_event.connect (() => {
```
Exalm, line 69: Interesting, so you basically check when the scale is pressed... I'm not sure there's a good way to replace that, maybe scale has something built-in. But I'd definitely not use any events nor GtkGesture here.

src/Widgets/SeekBar.vala
```diff
@@ -61,35 +61,35 @@ public class Music.SeekBar : Gtk.Grid {
...
 
         column_spacing = 6;
-        get_style_context ().add_class (Granite.STYLE_CLASS_SEEKBAR);
-        add (position_label);
-        add (scale);
-        add (duration_label);
+        css_classes = {"seekbar"};
+        attach (position_label, 0, 0);
+        attach (scale, 1, 0);
+        attach (duration_label, 2, 0);
```
Exalm, line 87: This is exactly what I meant with box:

```vala
append (position_label);
append (scale);
append (duration_label);
```

etc. And you have `prepend()` and `insert()` too. Grid is more flexible as it handles, well, grids, but is also way more complex to use.

Exalm (commented):
Some comments from the stream, some new ones. :)

src/MainWindow.vala
Exalm: FTR setting titlebar to an invisible widget is the only thing AdwApplicationWindow does now. And the only reason it exists.

src/MainWindow.vala
danrabbit: This is to make the entire background draggable so probably still want a handle here. Thanks!

src/Widgets/AlbumImage.vala
danrabbit: Yeah it used to be that they were saying Box was going to be removed in GTK4 so we were avoiding it. Box probably works here

Exalm: Nah, that was never the case. HBox and VBox were - and they _were_ removed, GtkBox was new in GTK3.

There were a few misleading statements in docs though, but those had been corrected since.

src/MainWindow.vala
Exalm: Ah ok, then you indeed need it. Just saw a few places with GtkHeaderBars within HdyWindowHandles in the past, so this looked like one of those places. :)

src/Views/NowPlayingView.vala
danrabbit: Yeah the problem with that is it doesn't seem there's a way to set icon size, so this seems the way for that unless we use CSS later

PR #695, opened 2022-01-21
PlaybackManager: shuffle

### LINKED ISSUE #687: Shuffle Queue
### Problem

There's currently no function to shuffle playback

### Proposal

The queue should always reflect the playback order, so a shuffle mechanism would need to rearrange the list itself, not just select a random track when clicking next. This means we'll probably need to keep two lists, one shuffled and one un-shuffled

### Prior Art (Optional)

_No response_

danrabbit:
Fixes #687 

Drafting because there's no "un-shuffle"


danrabbit:
It sounds like maybe we don't care much about "un-shuffle"? So hey let's go and we can always add that later if there's demand for it

PR #697, opened 2022-01-21
Move metadata discovery to AudioObject

### LINKED ISSUE #693: Move Metadata Discoverer to AudioObject?
Currently the metadata discoverer runs when adding files to the queue, which is fine for this design, but if there are future views, we probably want metadata even for audio objects that aren't queued.

Now that metadata discovery isn't attached to the playbin bus, we can probably move this to AudioObject so that it runs whenever an audio object is created. This would make all metadata storage pretty self contained there and keep the playback manager scoped to only handling playback and not also metadata

danrabbit:
Fixes #693 

* Discover metadata whenever an AudioObject is created
* PlaybackManager only handles playback now
* When discoverer returns metadata, more easily check which object it's for

### DIFF
--- a/src/AudioObject.vala
+++ b/src/AudioObject.vala
@@ -6,11 +6,122 @@
 public class Music.AudioObject : Object {
     public string uri { get; construct; }
     public Gdk.Texture texture { get; set; }
-    public string artist { get; set; }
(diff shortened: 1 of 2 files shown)


danrabbit:
Getting a bug here with playback position. Probably because track length is being discovered after we clamp position

PR #701, opened 2022-04-11
Create Flatpak deploy actions

danrabbit:
Adds Music daily builds to the AppCenter remote and stable build on release tag

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -1,0 +1,52 @@
+---
+
+name: Merge
+
+on:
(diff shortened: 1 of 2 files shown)


zeebok (approved):
I am not crazy familiar with writing actions for Flatpak stuff but of what I do know or can follow, this looks good. I assume when you are AppCenter remote that is referring to the daily build branch?

JoseExposito (approved):
LGTM. I compared both files with other projects and there are no surprises :)

PR #710, opened 2022-05-21
Use file basename instead of uri for the default audio object title.

vjr:
It appears that if the audio file doesn't have a title tag then Music uses the ugly file URI by default. This PR uses the prettier file basename if available otherwise falls back to the URI anyway.

See the track title on both panes - playlist on the left and above the progress bar on the right.

Before:

image

After:

image


danrabbit (approved):
Much better default, thanks!

PR #711, opened 2022-05-21
Skip invalid files instead of stopping playback.

vjr:
If you enqueue invalid files (such as I did when selecting two audio and one text file) then playback stops at the invalid file.

This PR simply skips to the next (or previous, if the previous button pressed) file instead of stopping at the invalid file.

Example screenshot:

image


vjr:
Also tested with multiple invalid files adjacent to each other and it skips over all of them to the next (or previous) valid audio file.

Marukesu:
i believe filtering the filechooser/drag'n'drop/`Application.open()` code to music files only an better solution. that will also stop the document portal of exporting unneeded/unused files.

vjr:
> i believe filtering the filechooser/drag'n'drop/`Application.open()` code to music files only an better solution. that will also stop the document portal of exporting unneeded/unused files.

Yes, that would be better. I'm hoping this PR can be a stop-gap merged until the proper excluding of invalid files is done?

ZenitsuDev:
In https://github.com/elementary/music/blob/main/src/MainWindow.vala#119 maybe we can put a function like GLib.ContentType.guess to get the mime type of the file dropped, and only append the file to the file array if it has audio mime type

ZenitsuDev:
As for the invalid files, maybe we can toast an error message?

Marukesu:
> I'm hoping this PR can be a stop-gap merged until the proper excluding of invalid files is done?

yeah, this PR is still valid even if the filtering is done, since there's can be errors about missing codecs or if the file was moved/deleted. one thing that i was think is, if it make sense to show an notification/toast when an error occurs.

vjr:
I'm completely useless when it comes to UI aspects haha so I welcome commits to this branch if we want the toast in this PR itself.

In the meantime I'll look into the suggestion about checking mime type...

ZenitsuDev:
@vjr maybe we can call a signal in the queue_files function in PlaybackManager if the mimetype of the dropped or 'opened by terminal' files are not audio mime types. Then connect the toast to it, then if the codecs of the file is missing, we skip it with this PR and call another toast to say that probably the codec for the audio is missing?

ZenitsuDev:
@vjr I am satisfied and I think we can mark it as ready for review😁

src/PlaybackManager.vala
```diff
@@ -109,8 +126,8 @@ public class Music.PlaybackManager : Object {
                     ngettext (
                         "%d track was added to the queue",
                         "%d tracks were added to the queue",
-                        files.length
-                    ).printf (files.length)
+                        files.length - invalids
```
danrabbit, line 129: This is probably a super edge case, but we should check here to make sure `files.length - invalids > 0` before sending a notification since it's possible now to get a notification that says "0 tracks were added to the queue"

src/PlaybackManager.vala
```diff
@@ -22,6 +23,14 @@ public class Music.PlaybackManager : Object {
     private uint progress_timer = 0;
     private Settings settings;
 
+    private enum Direction {
+        NONE,
+        NEXT,
+        PREVIOUS
+    }
+
+    private Direction direction = Direction.NONE;
```
danrabbit, line 32: This seems kind of fragile to me. Do we still need this since we're no longer adding the files into the queue? Is it safe to just assume we also go `next ()`?

danrabbit (dismissed):
Interaction here looks good. Nice work handling this garbage input! Good call on using a toast as well. Just a couple inline comments

src/PlaybackManager.vala
vjr: OK, I'll avoid showing a notification for "0 tracks added to Q".

src/PlaybackManager.vala
vjr: I think this might still be needed, because invalid tracks can still get added to the Q, I mean the `GLib.ContentType.guess ()` appears to just be looking at the file extension, I can still get a text file renamed as mp3 added to the Q. Which implies if there are corrupted audio tracks they would require this handling.

Also, people could click the previous button too to go back to earlier tracks, right?

vjr: I'll leave things here as is, or I could remove the `NONE` enum and initialise the var to `NEXT` instead?

PR #713, opened 2022-05-22
Use signal to notify when to show error toast

ZenitsuDev:
Aims to help with https://github.com/elementary/music/pull/711

@vjr I tried it :)

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -62,8 +62,15 @@ public class Music.MainWindow : Gtk.ApplicationWindow {
         queue.attach (scrolled, 0, 1);
         queue.add_controller (drop_target);
 
-        var queue_handle = new Gtk.WindowHandle () {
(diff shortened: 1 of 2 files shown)


vjr:
Thanks for this!

What if I import large number of files and there are many invalid files? Will there be a flood of toasts?

I'll test your branch later today , thanks again!

ZenitsuDev:
@vjr unfortunately yes at least in this way, they would appear one after another. maybe we can stockpile invalid filenames in an array and show them in one go.

vjr:
@ZenitsuDev  I made a change to only show one toast with the count of invalid files. If this part looks OK to you I can merge to my branch and we can continue working on that single branch for easier review from the devs.

ZenitsuDev:
@vjr yeah sure! I think just showing the numbers is good

Issue #721, opened 2022-07-22
music player themes?


mrlinuxdude:
### Problem

I'm using the new music player and really love it. Trying to get my favorite theme numix-gtk-blue on it doesn't look right and no icons show up

### Proposal

will other themes work out of the box with this music player 7. or is there option to downgrade to the older version if we want other themes then elementary's

danrabbit:
Music is built against the elementary stylesheet and icons and depends on the styles found there to look right. I've submitted a pull request to hard code the styles we require so that Music will still work correctly on non-elementary OS systems. It's not really feasible for us to support arbitrary stylesheets and icon libraries, sorry!

mrlinuxdude:
Thanks for the request

PR #723, opened 2022-08-08
Application: hard code gtk styles

### LINKED ISSUE #718: Adapt other gtk themes
### Problem

Only elementary themes and icons work with the new music player, If you switch to another theme outside of elementary app won't look as it should.

### Proposal

fix so you can add what ever theme you want or icons.

### Prior Art (Optional)

_No response_

danrabbit:
Fixes #718 #721 

We've received a few duplicate issue reports about styles being broken on non-elementary OS systems. This hard codes styles to use the elementary ones. Since the icons and stylesheet are already included in the Flatpak platform, we can assume they exist

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -74,6 +74,9 @@ public class Music.Application : Gtk.Application {
         var granite_settings = Granite.Settings.get_default ();
         var gtk_settings = Gtk.Settings.get_default ();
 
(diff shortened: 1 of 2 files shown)


jeremypw (approved):
This does not alter the appearance (although I am not sure why the accent color is hard coded as orange rather than following the system setting)

PR #724, opened 2022-08-24
Metainfo: Use branding tag and screenshot caption

danrabbit:
Works with https://github.com/elementary/appcenter/pull/1883

Also add `light` and `dark` colors for future use in AppCenter


jeremypw (approved):
Makes sense

PR #733, opened 2022-12-13
Work around play icon spinner sometimes not spinning.

vjr:
Fixes in three places:

* When you first open the app then select multiple tracks from the Files app the first track play icon spinner was stuck.
* When selecting different tracks in the playlist/queue the spinner would sometimes get stuck.
* When clicking the play/pause button repeatedly sometimes the spinner would get stuck.

### DIFF
--- a/src/PlaybackManager.vala
+++ b/src/PlaybackManager.vala
@@ -114,6 +114,7 @@ public class Music.PlaybackManager : Object {
         if (current_audio == null) {
             var audio_object = (AudioObject) queue_liststore.get_object (0);
             if (audio_object != null) {
+                // Timeout is a workaround for https://gitlab.gnome.org/GNOME/gtk/-/issues/4426
                 GLib.Timeout.add (250, () => {
                     current_audio = audio_object;
                     return Source.REMOVE;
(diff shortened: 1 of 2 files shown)


danirabbit (dismissed):
I'm not a huge fan of adding timeouts with magic numbers, but can we at least add a comment that this is a workaround and a link to the upstream issue: https://gitlab.gnome.org/GNOME/gtk/-/issues/4426

vjr:
> I'm not a huge fan of adding timeouts with magic numbers, but can we at least add a comment that this is a workaround and a link to the upstream issue: https://gitlab.gnome.org/GNOME/gtk/-/issues/4426

The 250 ms timeout could be an issue if there are tens of thousands of tracks (`TrackRow`s) each connecting to the `current-audio` property notification of the PlaybackManager?

zeebok:
Looks like the upstream has a fix in? If so, this should be re-evaluated.

danirabbit:
Yeah this was recently fixed upstream so I'm not sure it's worth carrying this workaround when a platform update will solve it automatically

vjr:
Oh good to hear it's fixed upstream. Do we want to close this PR right away or wait until the fix actually lands in an update?

danirabbit:
Yeah, I feel like we should probably just wait for a platform update with a newer Gtk release and not carry the workaround since it's a pretty minor bug

PR #747, opened 2023-04-01
Allow to open directories directly

### LINKED ISSUE #709: Open entire folder from Files
### Problem

Currently it is not possible to right click on a folder in Files and open it in Music, so that all audio files in that folder get put into the playlist. You have to enter the folder, select all files and then open them in with Music which seem like an overcomplicated way of doing that.

### Proposal

Make it possible to just open an entire folder with Music and have all audio files in that folder appear in the queue.

### Prior Art (Optional)

_No response_

### LINKED ISSUE #716: Open song or folder from Music
### Problem

Music doesn't have an option to open song or folder directly from the app.

### Proposal

Have option to open File or Folder directly from Music.

### Prior Art (Optional)

_No response_

aitor-gomila:
- [x] Add `inode/directory` MIME type to .desktop file.
- [x] Support opening directories from code.
- [x] Support opening folders from file explorer

These will be worked on a separate PR:
- ~~[ ] Add `audio/x-mpegurl` MIME type to .desktop file.~~
- ~~[ ] Support opening playlist files from code.~~

Fixes #709

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -116,8 +116,8 @@ public class Music.Application : Gtk.Application {
         File[] elements = {};
 
         while ((name = dir.read_name ()) != null) {
-            string file_path = Path.build_filename (directory, name);
-            File file = File.new_for_path (file_path);
+            var file_path = Path.build_filename (directory, name);
+            var file = File.new_for_path (file_path);
 
             elements += file;
         }
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -103,7 +103,7 @@ public class Music.Application : Gtk.Application {
         settings.bind ("window-maximized", main_window, "maximized", SettingsBindFlags.SET);
     }
 
-    File[] listDirectory(string directory) {
+    File[] list_directory(string directory) {
         var dir = Dir.open(directory, 0);
 
         string? name = null;
@@ -119,30 +119,28 @@ public class Music.Application : Gtk.Application {
         return elements;
     }
 
-    private bool isDirectory(string file) {
+    private bool is_directory(string file) {
         return FileUtils.test(file, FileTest.IS_DIR);
     }
 
-    private void loopThroughFiles(File[] files) {
-        // All of these are added later in bulk
+    private File[] loop_through_files(File[] files) {
+        // All of these will be returned later in bulk
         File[] elements = {};
 
         foreach (var file in files) {
             var filePath = file.get_path();
-            var isDirectory = isDirectory(filePath);
+            var isDirectory = is_directory(filePath);
 
-            if (!isDirectory) {
-                elements += file;
-
+            if (isDirectory) {
+                var directoryElements = list_directory(filePath);
+                elements += loop_through_files(directoryElements);
                 continue;
             }
 
-            // Is a directory
-            var directoryElements = listDirectory(filePath);
-            loopThroughFiles(directoryElements);
+            elements += file;
         }
 
-        playback_manager.queue_files (elements);
+        return elements;
     }
 
     protected override void open (File[] files, string hint) {
@@ -150,7 +148,7 @@ public class Music.Application : Gtk.Application {
             activate ();
         }
 
-        loopThroughFiles(files);
+        playback_manager.queue_files(loop_through_files(files));
     }
 
     private void action_play_pause () {
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -103,45 +103,45 @@ public class Music.Application : Gtk.Application {
         settings.bind ("window-maximized", main_window, "maximized", SettingsBindFlags.SET);
     }
 
-    File[] list_directory(string directory) {
+    File[] list_directory (string directory) {
         try {
-            var dir = Dir.open(directory, 0);
+            var dir = Dir.open (directory, 0);
 
             string? name = null;
             File[] elements = {};
 
-            while((name = dir.read_name ()) != null) {
-                string filePath = Path.build_filename (directory, name);
-                File file = File.new_for_path(filePath);
+            while ((name = dir.read_name ()) != null) {
+                string file_path = Path.build_filename (directory, name);
+                File file = File.new_for_path (file_path);
 
                 elements += file;
             }
 
             return elements;
         } catch (FileError e) {
-            warning(e.message);
+            warning (e.message);
             return {};
         }
     }
 
-    private bool is_directory(string file) {
-        return FileUtils.test(file, FileTest.IS_DIR);
+    private bool is_directory (string file) {
+        return FileUtils.test (file, FileTest.IS_DIR);
     }
 
-    private File[] loop_through_files(File[] files) {
+    private File[] loop_through_files (File[] files) {
         // All of these will be returned later in bulk
         File[] elements = {};
 
         foreach (var file in files) {
-            var filePath = file.get_path();
-            var isDirectory = is_directory(filePath);
+            var file_path = file.get_path ();
+            var is_directory = is_directory (file_path);
 
-            if (isDirectory) {
-                var directoryElements = list_directory(filePath);
-                var directoryFiles =  loop_through_files(directoryElements);
+            if (is_directory) {
+                var directory_elements = list_directory (file_path);
+                var directory_files = loop_through_files (directory_elements);
 
-                foreach (var directoryFile in directoryFiles) {
-                    elements += directoryFile;
+                foreach (var directory_file in directory_files) {
+                    elements += directory_file;
                 }
 
                 continue;
@@ -158,7 +158,7 @@ public class Music.Application : Gtk.Application {
             activate ();
         }
 
-        playback_manager.queue_files(loop_through_files(files));
+        playback_manager.queue_files (loop_through_files (files));
     }
 
     private void action_play_pause () {
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -104,19 +104,24 @@ public class Music.Application : Gtk.Application {
     }
 
     File[] list_directory(string directory) {
-        var dir = Dir.open(directory, 0);
+        try {
+            var dir = Dir.open(directory, 0);
 
-        string? name = null;
-        File[] elements = {};
+            string? name = null;
+            File[] elements = {};
 
-        while((name = dir.read_name ()) != null) {
-            string filePath = Path.build_filename (directory, name);
-            File file = File.new_for_path(filePath);
+            while((name = dir.read_name ()) != null) {
+                string filePath = Path.build_filename (directory, name);
+                File file = File.new_for_path(filePath);
 
-            elements += file;
+                elements += file;
+            }
+
+            return elements;
+        } catch (FileError e) {
+            warning(e.message);
+            return {};
         }
-
-        return elements;
     }
 
     private bool is_directory(string file) {
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -104,8 +104,9 @@ public class Music.Application : Gtk.Application {
     }
 
     private File[] list_directory (string directory) {
+        Dir dir;
         try {
-            var dir = Dir.open (directory, 0);
+            dir = Dir.open (directory, 0);
         } catch (FileError e) {
             warning (e.message);
 
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -109,7 +109,6 @@ public class Music.Application : Gtk.Application {
             dir = Dir.open (directory, 0);
         } catch (FileError e) {
             warning (e.message);
-
             return {};
         }
 
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -124,10 +124,6 @@ public class Music.Application : Gtk.Application {
         }
     }
 
-    private bool is_directory (string file) {
-        return FileUtils.test (file, FileTest.IS_DIR);
-    }
-
     private File[] loop_through_files (File[] files) {
         // All of these will be returned later in bulk
         File[] elements = {};
@@ -135,7 +131,7 @@ public class Music.Application : Gtk.Application {
         foreach (var file in files) {
             var file_path = file.get_path ();
 
-            if (is_directory (file_path)) {
+            if (FileUtils.test (file_path, FileTest.IS_DIR)) {
                 var directory_elements = list_directory (file_path);
                 var directory_files = loop_through_files (directory_elements);
 
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -134,9 +134,8 @@ public class Music.Application : Gtk.Application {
 
         foreach (var file in files) {
             var file_path = file.get_path ();
-            var is_directory = is_directory (file_path);
 
-            if (is_directory) {
+            if (is_directory (file_path)) {
                 var directory_elements = list_directory (file_path);
                 var directory_files = loop_through_files (directory_elements);
 
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -124,15 +124,15 @@ public class Music.Application : Gtk.Application {
     }
 
     private void loopThroughFiles(File[] files) {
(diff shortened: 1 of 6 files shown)


lenemter (commented):
When I launch Music from terminal using `./src/io.elementary.music /home/lenemter/Music/`, it only opens 1 file.

aitor-gomila:
Got it, will check it out again ASAP. Thanks for taking the time to review my code!

src/Application.vala
```diff
@@ -103,12 +103,57 @@ public class Music.Application : Gtk.Application {
         settings.bind ("window-maximized", main_window, "maximized", SettingsBindFlags.SET);
     }
 
+    // FIXME(aitor-gomila): improve algorithm and make code cleaner
+    Array<File> listDirectory(string directory) {
+        var dir = Dir.open(directory, 0);
+        string? name = null;
+        Array<File> elements = new Array<File>();
```
lenemter, line 110: Why not use File[]?

src/Application.vala
```diff
@@ -103,12 +103,57 @@ public class Music.Application : Gtk.Application {
...
+        for (var i = 0; i < files.length; i++) {
+            var file = files.index(i);
+            var filePath = file.get_path();
+            var isDirectory = isDirectory(filePath);
+
+            if (!isDirectory) {
+                File[] arr = new File[1];
+                arr[0] = file;
+                playback_manager.queue_files(arr);
+                return;
```
lenemter, line 134: You shouldn't return here. User may have opened many files, not just one.

src/Application.vala
```diff
@@ -103,12 +103,57 @@ public class Music.Application : Gtk.Application {
...
+
     protected override void open (File[] files, string hint) {
         if (active_window == null) {
             activate ();
         }
 
-        playback_manager.queue_files (files);
+        // Turns File[] into Array<Files>
+        // Is there a cleaner way to do this?
+        Array<File> filesArr = new Array<File>();
```
lenemter, line 149: Why not use File[]?

lenemter (changes requested):
I left some inline comments. Take a look at them please.

Also check out https://docs.elementary.io/develop/writing-apps/code-style to match all the other code here.

src/Application.vala
aitor-gomila: Native C arrays need to have a set size. So if you want to open a directory inside a directory, you need to get the number of files in that folder, or just use a GLib Array. Probably better to use native C arrays in this case, will check out how it's done.

src/Application.vala
aitor-gomila: The loop should go through all files, return should only stop the current iteration for the current file. What should I use instead?

src/Application.vala
lenemter: They are resizable in Vala https://wiki.gnome.org/Projects/Vala/Tutorial#Arrays

src/Application.vala
lenemter: Use `continue`

src/Application.vala
aitor-gomila: Is there a way to get the quantity of files in a directory beforehand?

aitor-gomila:
Ok, resolved

src/Application.vala
```diff
@@ -103,12 +103,48 @@ public class Music.Application : Gtk.Application {
         settings.bind ("window-maximized", main_window, "maximized", SettingsBindFlags.SET);
     }
 
+    // FIXME(aitor-gomila): improve algorithm and make code cleaner
+    File[] listDirectory(string directory) {
+        var dir = Dir.open(directory, 0);
+        string? name = null;
+        File[] elements = new File[1];
```
lenemter, line 110:
```suggestion
        File[] elements = {};
```

src/Application.vala
```diff
@@ -103,12 +103,48 @@ public class Music.Application : Gtk.Application {
...
+
+        while((name = dir.read_name ()) != null) {
+            string filePath = Path.build_filename (directory, name);
+            File file = File.new_for_path(filePath);
+            elements += file;
+        }
+        return elements;
```
lenemter, line 117: Add a blank line  before return please, just for the sake of readability.

src/Application.vala
```diff
@@ -103,12 +103,48 @@ public class Music.Application : Gtk.Application {
...
+    }
+
+    private bool isDirectory(string file) {
+        return FileUtils.test(file, FileTest.IS_DIR);
+    }
+
+    private void loopThroughFiles(File[] files) {
+        for (var i = 0; i < files.length; i++) {
+            var file = files[i];
```
lenemter, line 126:
```suggestion
        foreach (var file in files) {
```

lenemter (changes requested):

lenemter:
> Ok, resolved. How do I lint the code? Ran what's on the CI, but it didn't work (vala-lint is not installed in my machine).

vala-lint unfortunately doesn't automatically lint the code, it only shows lint errors. If you want you can install it from https://github.com/vala-lang/vala-lint

BAProductions:
> > Ok, resolved. How do I lint the code? Ran what's on the CI, but it didn't work (vala-lint is not installed in my machine).
> 
> 
> 
> vala-lint unfortunately doesn't automatically lint the code, it only shows lint errors. If you want you can install it from https://github.com/vala-lang/vala-lint

Yes it does An I've done it

lenemter:
Oh I didn't know there is a `--fix` flag. Nice discovery.

aitor-gomila:
I don't have my PC with me right now. Do the new changes work? Can someome verify that?

src/Application.vala
aitor-gomila: Nice.

src/Application.vala
```diff
@@ -103,12 +103,52 @@ public class Music.Application : Gtk.Application {
         settings.bind ("window-maximized", main_window, "maximized", SettingsBindFlags.SET);
     }
 
+    File[] listDirectory(string directory) {
```
lenemter, line 106:
```suggestion
    File[] list_directory (string directory) {
```

src/Application.vala
```diff
@@ -103,12 +103,52 @@ public class Music.Application : Gtk.Application {
...
+        }
+
+        return elements;
+    }
+
+    private bool isDirectory(string file) {
+        return FileUtils.test(file, FileTest.IS_DIR);
+    }
+
+    private void loopThroughFiles(File[] files) {
```
lenemter, line 126:
```suggestion
    private void loop_through_files (File[] files) {
```

src/Application.vala
```diff
@@ -103,12 +103,52 @@ public class Music.Application : Gtk.Application {
...
+        foreach (var file in files) {
+            var filePath = file.get_path();
+            var isDirectory = isDirectory(filePath);
+
+            if (!isDirectory) {
```
lenemter, line 134: Can you please reverse this if? Like check if file is directory and then handle it as a directory rather than checking if it's a file

src/Application.vala
```diff
@@ -103,12 +103,52 @@ public class Music.Application : Gtk.Application {
...
+                File[] arr = {};
+                arr[0] = file;
```
lenemter:
```suggestion
                File[] arr = { file };
```

lenemter (changes requested):

src/Application.vala
```diff
@@ -103,12 +103,54 @@ public class Music.Application : Gtk.Application {
...
+
+                continue;
+            }
+
+            // Is a directory
+            var directoryElements = listDirectory(filePath);
+            loopThroughFiles(directoryElements);
+        }
+
+        playback_manager.queue_files (elements);
```
lenemter, line 145: Nice work, this is better now then adding files one by one. But i think it would be even better if `loop_through_files` will return array of files and then they will be added to queue in `open` method.

aitor-gomila: Is there a way to move the elements of an array to another in vala? E.g. move `[1, 2, 3]` to array `[4, 5, 6]`. Result: `[4, 5, 6, 1, 2, 3]` not: `[4, 5, 6, [1, 2, 3]]`

lenemter: You can add them in a loop

src/Application.vala
```diff
@@ -103,12 +103,52 @@ public class Music.Application : Gtk.Application {
         settings.bind ("window-maximized", main_window, "maximized", SettingsBindFlags.SET);
     }
 
+    File[] listDirectory(string directory) {
+        var dir = Dir.open(directory, 0);
```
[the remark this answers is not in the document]
aitor-gomila: Try/catch?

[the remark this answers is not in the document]
lenemter: Yep

src/Application.vala
aitor-gomila: What do you mean? Check my new code at [line 136](https://github.com/elementary/music/pull/747/commits/a049712dd1446e63fb68a74cf89afaeea49bb8d9#diff-f4e5d48d751c670a8575715da1cb0a0976532427a8b6143c7fdede415a33ba6cR136)

lenemter: Oh it's fine now, sorry

lenemter: You need to look through loop_through_files again and add elements one by one

src/Application.vala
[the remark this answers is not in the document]
aitor-gomila: How do you log errors softly?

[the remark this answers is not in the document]
lenemter: Use `warning (e.mesaage)`

aitor-gomila:
Fixed the style of code. For some reason, it works from the terminal but not from the file explorer: https://youtu.be/pNTw4lanGbo

src/Application.vala
```diff
@@ -103,12 +103,62 @@ public class Music.Application : Gtk.Application {
...
+            }
+
+            return elements;
+        } catch (FileError e) {
+            warning (e.message);
+            return {};
+        }
+    }
+
+    private bool is_directory (string file) {
```
jeremypw, line 127: This is only used in one place so can be inlined.

lenemter:
> Fixed the style of code. For some reason, it works from the terminal but not from the file explorer: https://youtu.be/pNTw4lanGbo

That's happening because in this case it launches flatpak version of Music. You need to delete it through AppCenter or through Terminal by executing `flatpak remove io.elementary.Music` and then it should work.

aitor-gomila:
Ok, need to make the file methods private and make one function inline and then it should be good to go.

src/Application.vala
```diff
@@ -103,12 +103,61 @@ public class Music.Application : Gtk.Application {
...
+    }
+
+    private File[] loop_through_files (File[] files) {
+        // All of these will be returned later in bulk
+        File[] elements = {};
+
+        foreach (var file in files) {
+            var file_path = file.get_path ();
+
+            if (is_directory (file_path)) {
```
lenemter, line 138:
```suggestion
            if (FileUtils.test (file_path, FileTest.IS_DIR)) {
```

src/Application.vala
```diff
@@ -103,12 +103,57 @@ public class Music.Application : Gtk.Application {
...
+
+            while ((name = dir.read_name ()) != null) {
+                string file_path = Path.build_filename (directory, name);
+                File file = File.new_for_path (file_path);
+
+                elements += file;
+            }
+
+            return elements;
+        } catch (FileError e) {
```
lenemter, line 121: Can you put only `var dir = Dir.open (directory, 0);` in try-catch please?

aitor-gomila: Won't everything else execute then anyway if something goes wrong?

aitor-gomila: Done

src/Application.vala
```diff
@@ -103,12 +103,58 @@ public class Music.Application : Gtk.Application {
         settings.bind ("window-maximized", main_window, "maximized", SettingsBindFlags.SET);
     }
 
+    private File[] list_directory (string directory) {
+        try {
+            var dir = Dir.open (directory, 0);
```
lenemter, line 108:
```suggestion
        Dir dir;
        try {
            dir = Dir.open (directory, 0);
```

lenemter (approved):
Code looks good. I'll *try* to package music as flatpak locally and see if it works.

aitor-gomila:
The only thing that doesn't work for me is that by default, Music doesn't open folders, even though I set `inode/directory`. If set manually it works fine, though

lenemter:
Well, it doesn't work :(
@danirabbit You know more about flatpak, so maybe you have an idea why?

aitor-gomila:
Isn't it expected that the app needs access to the filesystem? It should be on by default though.

jeremypw:
This works when run natively but not when in a Flatpak (but master doesn't anyway).  I find Files does give Music as one of the options in the "Open With" submenu.  I am working on the Flatpak version atm.  You can debug the program while running in a Flatpak by launching with `flatpak run --command=sh io.elementary.music`. This gives you a shell prompt inside the sandbox from where you can launch the program as usual but can see all the debugging and error messages as well as being able to run `gdb` if you need to.

jeremypw:
@aitor-gomila I've pushed a PR to your fork with a small fix for Flatpak compatibility.  With this I can open files and folders from Files with Music both by using the context menu and by drag-drop.  Launching from terminal with `flatpak run io.elementary.music <path to file>` also works.  All providing that the folders/files are withing `~/Music` of course.  (or whatever XDG_MUSIC is pointing to)

aitor-gomila:
Thanks. How did Music work without the changes on your PR?

jeremypw:
Note that giving access to the home directory with `filesystem=home` does *not* give access to all its subfolders for some reason. But good practice is to allow each app access only to its one folder (in this case Music)

src/Application.vala
```diff
@@ -103,12 +103,61 @@ public class Music.Application : Gtk.Application {
         settings.bind ("window-maximized", main_window, "maximized", SettingsBindFlags.SET);
     }
 
+    private File[] list_directory (string directory) {
+        Dir dir;
+        try {
+            dir = Dir.open (directory, 0);
+        } catch (FileError e) {
+            warning (e.message);
+
```
jeremypw, line 112: Unneeded empty line

src/Application.vala
```diff
@@ -103,12 +103,61 @@ public class Music.Application : Gtk.Application {
...
+            return {};
+        }
+
+        string? name = null;
+        File[] elements = {};
+
+        while ((name = dir.read_name ()) != null) {
+            string file_path = Path.build_filename (directory, name);
+            File file = File.new_for_path (file_path);
```
jeremypw, line 121:
```suggestion
            var file = File.new_for_path (file_path);
```

jeremypw (changes requested):
A couple of minor codestyle things. Then it will be OK with me.

src/Application.vala
aitor-gomila: Same for file_path right?

aitor-gomila:
Done. Thanks for taking the time to review my PR

jeremypw:
One extra thing - the appdata for the next release should be updated with the new functionality.

aitor-gomila:
Appdata?

aitor-gomila:
misclick, sorry

jeremypw:
Sorry - its called metainfo in this repository I see.  Its the file `/data/music.metainfo.xml.in`.  You will see other <release></release> sections in there the last release was 7.0.1 in December so the new info needs to be added to the pending 7.1.  Looks like drag and drop is already entered so the new functionality is to open folders in Music from Files.  Don't worry - its not too important - it will be checked and updated as needed before the release is made anyway.

aitor-gomila:
@lenemter Can you check if `sudo make install` installs system-wide or just for the current user? I created a separate account and I'm able to open folders.

PR #751, opened 2023-04-04
Update metainfo for next release

jeremypw:
Adds info relating to last commit.

In passing adds another directory to `.gitignore`


danirabbit:
I think we should make sure to say "queue" here and not "playlist" since that's a specific name for a feature that isn't implemented yet

danirabbit (approved):
Looks good 👍 Ty!

PR #765, opened 2023-12-14
Implement simple search in playlist

### LINKED ISSUE #762: Search mechanism in the app
### Problem

We can't search for a song directly in the app in a playlist.

### Proposal

Build a search option in the header bar that allows you to search for a song in a playlist.

### Prior Art (Optional)

searchbar

jeremypw:
Fixes #762

 - Just adds a search entry to the queue header that appears whenever there are items in the queue. 
 - Entering a search term does a case insensitive search for the term anywhere in the titles in list and selects the first matching row (if any). 
 - If no match the currently selected item remains selected.  
 - The matching row does not play until the search entry is activated or the row is clicked on.  
 - The queue scrolls to show the matched item.
 - `<Ctrl>F` focuses the search entry if showing

Screenshot from 2023-12-14 16 49 39


zeebok:
This never seems to allow the first item in the list as a result to highlight, which is handy if I search in a list that is long and scrollable

jeremypw:
@zeebok Sorry for the delay in getting back to this.  I have now fixed the issue you noticed.

zeebok (approved):

PR #766, opened 2023-12-14
Add music file[s] from within app

jeremypw:
Towards a fix of #716 

 - Adds an actionbar/button linked to an open file[s] action.  
 - Open file[s] action shows a filechooser dialog for selecting multiple files
 - Selected files are added to the playlist.

This PR does not implement the open folder element of the issue as that would require a separate action/button. That can be simply added later if this PR is accepted.


src/Application.vala
```diff
@@ -175,6 +177,51 @@ public class Music.Application : Gtk.Application {
...
+        );
+        file_chooser.add_filter (music_files_filter);
+        file_chooser.add_filter (all_files_filter);
+        file_chooser.select_multiple = true;
+
+        file_chooser.response.connect ((response) => {
+            SList<weak File> file_list = null;
+            if (response == Gtk.ResponseType.ACCEPT) {
+                var files = file_chooser.get_files ();
+                File? file;
```
ryonakano, line 209: This variable seems to be unused:

```
../src/Application.vala:209.23-209.26: warning: Local variable `file' declared but never used
  209 |                 File? file;
      |                       ^~~~ 
```

src/Application.vala
```diff
@@ -175,6 +177,51 @@ public class Music.Application : Gtk.Application {
...
+        var music_files_filter = new Gtk.FileFilter () {
+            name = _("Music Files"),
+        };
+        music_files_filter.add_mime_type ("audio/*");
+
+        var filter_model = new ListStore (typeof (Gtk.FileFilter));
+        filter_model.append (all_files_filter);
+        filter_model.append (music_files_filter);
+
+        var file_chooser = new Gtk.FileChooserNative (
```
ryonakano, line 194: Gtk.FileChooserNative is deprecated since 4.10, so it would be better to avoid to use in new codes.

src/Application.vala
```diff
@@ -175,6 +177,51 @@ public class Music.Application : Gtk.Application {
...
+        var all_files_filter = new Gtk.FileFilter () {
+            name = _("All files"),
+        };
+        all_files_filter.add_pattern ("*");
+        var music_files_filter = new Gtk.FileFilter () {
+            name = _("Music Files"),
+        };
+        music_files_filter.add_mime_type ("audio/*");
+
+        var filter_model = new ListStore (typeof (Gtk.FileFilter));
```
ryonakano, line 190: What is this ListStore for? It looks like unused.

src/MainWindow.vala
```diff
@@ -161,6 +161,25 @@ public class Music.MainWindow : Gtk.ApplicationWindow {
...
+            }
+        }
+
+        file_list.reverse ();
+        foreach (unowned var file in file_list) {
+            file_array += file;
+        }
+
+        PlaybackManager.get_default ().queue_files (file_array);
+    }
```
ryonakano, line 182:
```suggestion
    }

```

We should have a newline here as the end of the method.

src/Application.vala
```diff
@@ -175,6 +177,51 @@ public class Music.Application : Gtk.Application {
         playback_manager.shuffle ();
     }
 
+    private void action_open () {
+        var all_files_filter = new Gtk.FileFilter () {
+            name = _("All files"),
+        };
+        all_files_filter.add_pattern ("*");
+        var music_files_filter = new Gtk.FileFilter () {
+            name = _("Music Files"),
```
ryonakano, line 186: `All files` (Sentence case) v.s. `Music Files` (Title Case)―maybe we should unify to either one.

ryonakano (changes requested):
Works fine, just some nitpicking comments.

PR #769, opened 2024-03-20
Application: use modern features for loading assets, use startup

danirabbit:


src/Application.vala
```diff
@@ -47,6 +46,29 @@ public class Music.Application : Gtk.Application {
...
+        granite_settings.notify["prefers-color-scheme"].connect (() => {
+            gtk_settings.gtk_application_prefer_dark_theme = (
+                granite_settings.prefers_color_scheme == DARK
+            );
+        });
+    }
+
+    protected override void activate () {
+        if (active_window != null) {
+            active_window.present_with_time (Gdk.CURRENT_TIME);
```
Marukesu, line 68:
```suggestion
            active_window.present ();
```

present_with_time() is never the right choice when there isn't a timestamp at hand.

in this case in particular Gtk.Application already handle the timestamp for us, so present() should work fine.

Marukesu (approved):

PR #793, opened 2025-03-17
Handle metadata discovery at creation of AudioObjects

### LINKED ISSUE #693: Move Metadata Discoverer to AudioObject?

### LINKED ISSUE #775: Re-adding a file causes a partially corrupt queue item
### What Happened?

I had accidentally dragged a single song instead of a whole album into Music. I figured I could just select all and drag into Music without issue but it created a messed up entry for the first song I had added.

Screenshot from 2024-05-24 12 44 10

This strange duplicate **does** play but the track progress bar doesn't work right, and obviously the metadata in the queue list is just the filename.

### Steps to Reproduce

1. Add a song
2. Add more music, including that first song again

### Expected Behavior

I would expect that the queue list would see the song (likely by filename) already in the list and not add it.

### OS Version

7.x (Horus)

### Software Version

Latest release (I have run all updates)

### Log Output

_No response_

### Hardware Info

_No response_

Oowoosh0:
Fixes #693 
Fixes #775 
Closes #697 

- Metadata discovery is handled when a new AudioObject is created.
- The discovery is delegated to a MetadataDiscoverer to avoid creating a new gstreamer discoverer for every AudioObject and to map the returned metadata back to the corresponding AudioObjects.
- This also fixes a bug where adding the same file multiple times led metadata not being updated for that file.

### DIFF
--- a/src/AudioObject.vala
+++ b/src/AudioObject.vala
@@ -9,8 +9,88 @@ public class Music.AudioObject : Object {
     public string artist { get; set; }
     public string title { get; set; }
     public int64 duration { get; set; default = 0; }
+    private static MetadataDiscoverer discoverer = new MetadataDiscoverer();
 
     public AudioObject (string uri) {
-        Object (uri: uri);
+        Object (
+            uri: uri,
+            title: uri
+        );
+    }
+
+    construct {
+        discoverer.request (this);
+    }
+
+    public void update_metadata (Gst.PbUtils.DiscovererInfo info) {
+        duration = (int64) info.get_duration ();
+
+        unowned Gst.TagList? tag_list = info.get_tags ();
+
+        string _title;
+        tag_list.get_string (Gst.Tags.TITLE, out _title);
+        if (_title != null) {
+            title = _title;
+        }
+
+        string _artist;
+        tag_list.get_string (Gst.Tags.ARTIST, out _artist);
+        if (_artist != null) {
+            artist = _artist;
+        } else if (_title != null) { // Don't set artist for files without tags
+            artist = _("Unknown");
+        }
+
+        var sample = get_cover_sample (tag_list);
+        if (sample != null) {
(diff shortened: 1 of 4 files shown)


danirabbit:
@Oowoosh0 can you resolve conflicts here please? :)

Oowoosh0:
rebased my branch conflicts should be gone now.

zeebok:
New conflicts!

src/AudioObject.vala
```diff
@@ -9,8 +9,88 @@ public class Music.AudioObject : Object {
...
+
+        var sample = get_cover_sample (tag_list);
+        if (sample != null) {
+            var buffer = sample.get_buffer ();
+
+            if (buffer != null) {
+                texture = Gdk.Texture.for_pixbuf (get_pixbuf_from_buffer (buffer));
+            }
+        }
+    }
```
danirabbit, line 52: These properties are all publicly settable so why do this here instead of in the MetadataDiscoverer?

danirabbit (commented):
Definitely cleans up the PlaybackManager. Adding a MetadataDiscoverer singleton makes sense to me. Just one question

src/AudioObject.vala
Oowoosh0: Yes this probably should be moved into the MetadataDiscoverer. Sorry I didn't have much time the last 2 weeks, but I'll have a lot more time from wednesday onward to update this.

zeebok:
New conflict to resolve @Oowoosh0

Issue #779, opened 2024-11-06
Make Queue editable


AustrianLinuxMemer:
### Problem

The problem I see with the queue as it stands now is that there is no way to edit the queue after files have been added. It is possible to add files to the queue but there is no way to move items in the queue or to clear the queue without closing the program

### Proposal

I would propose:
- To add a way to move the items in the queue per drag-and-drop
- To add a way to be able to remove items from the queue (either through some button on the item itself on the item or through a context menu)
- To add a way to be able to empty the queue without closing the application

### Prior Art (Optional)

![Unbenannt 1](https://github.com/user-attachments/assets/fe6744ce-6496-4838-a2ca-4732c5643d98)

Oowoosh0:
Would it be better to put the remove action into a context menu that appears on right clicking the item or to add a remove button to the end of each item that only shows up if you hover above the item?

danirabbit:
@Oowoosh0 hover actions don't show on touch, so it would probably be better to add a context menu

PR #795, opened 2025-03-19
Add clear queue button

Oowoosh0:
Adds a button to remove all elements from the queue.
One of the features requested in #779 
music-clear-queue

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -10,6 +10,7 @@ public class Music.Application : Gtk.Application {
     public const string ACTION_PREVIOUS = "action-previous";
     public const string ACTION_SHUFFLE = "action-shuffle";
     public const string ACTION_FIND = "action-find";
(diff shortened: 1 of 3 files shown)


Oowoosh0:
Just noticed that it would make sense to hide the button if the queue is already empty. I'm doing to add that.

danirabbit:
> hide the button if the queue is already empty

@Oowoosh0 Instead of this, you should disable the action and it'll automatically make the button insensitive and ignore any future keyboard  shortcuts :)

Oowoosh0:
the action is now disabled when no items are in the queue.
The disabled button is automatically slightly darker and has an inset box-shadow. Even though the style class is flat. I don't know if that is intended by the elementary stylesheet or if that is a bug.
music-clear-2

danirabbit (approved):
Looks good!

I see what you mean with the styling. That's definitely a stylesheet bug and not anything for you to fix here

PR #796, opened 2025-03-19
MainWindow: replace ListBox with ListView

danirabbit:
Can be rebase merged

### DIFF
--- a/src/Widgets/TrackRow.vala
+++ b/src/Widgets/TrackRow.vala
@@ -12,9 +12,9 @@ public class Music.TrackRow : Granite.Bin {
 
         set {
             if (_audio_object != null) {
-                artist_binding.unbind ();
-                texture_binding.unbind ();
-                title_binding.unbind ();
+                _audio_object.notify["artist"].disconnect (update_artist_label);
+                _audio_object.notify["title"].disconnect (update_title_label);
+                _audio_object.notify["texture"].disconnect (update_cover_art);
             }
 
             _audio_object = value;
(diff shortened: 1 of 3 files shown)


src/Widgets/TrackRow.vala
```diff
@@ -4,37 +4,31 @@
  */
 
 public class Music.TrackRow : Gtk.ListBoxRow {
```
Oowoosh0, line 6: With the change to ListView TrackRow shouldn't inherit from ListBoxRow any more. Leads to some runtime errors when grabbing focus.

Oowoosh0 (changes requested):

src/Widgets/TrackRow.vala
danirabbit: Woops thank you for catching that

Oowoosh0:
Sorry there is one more thing that I hadn't noticed before, but this leaks memory.
New TrackRows are constantly created as you scroll the ListView up or down, but they never get destroyed. They also don't get destroyed when emptying the list. So memory consumption just slowly goes up as the app is used. We probably need to use the unbind or teardown signal from the SignalListItemFactory to handle that.

danirabbit:
@Oowoosh0 good catch! I'm still learning ListView but that makes a lot of sense. I think it's working as expected now. If not, let me know how you're testing and I'll keep poking at it :)

Oowoosh0:
I used the very professional method of opening a system monitor, opening over 700 music files in the app and then crazily scrolling up and down. Which makes the memory usage slowly creep up over time. It does not do so in the version still using ListBox. Although TrackRows seems to be destroyed correctly actually.
I'll look more into it, but we could also still go ahead and merge it since it only goes up slowly and you would need to crazily scroll around for a long time before it becomes a problem.

Oowoosh0:
I think it is the Bindings. New Bindings are being created whenever a different AudioObject is set in a TrackRow but they never get destroyed. I thought unbind() would take care of that but for some reason it does not seem to do so :/

Oowoosh0:
@danirabbit connecting and disconnecting to the notify signal manually and not using bind_property actually gets rid of the leak. Calling unbind() for some reason does not free the memory used by a Binding and bind_property creates new Bindings every time a new AudioObject is set to a TrackRow. If you haven't done any more work on this branch then I can push that change.

danirabbit:
@Oowoosh0 sure go ahead :) Thanks!

PR #797, opened 2025-03-23
Remove songs

### LINKED ISSUE #787: Ability to remove a track
### Problem

From my understanding, this is trying to be a queued based VLC like app rather than a music app which retains song information. If so, the ability to remove a track after adding it would be useful

### Proposal

Right click to remove or press delete button

### Prior Art (Optional)

_No response_

Oowoosh0:
This adds the ability to remove single items from the queue, either by right clicking on the item or by using the "delete" keyboard shortcut on the selected item.
Fixes #787 
and implements another feature of #779 

Screenshot from 2025-03-23 10 09 40

Currently right clicking the song doesn't select it. This can make it hard to know which song will be affected by the action if you clicked somewhere in between two songs, because there is no visual cue to tell which song is selected. But I think that will be easier to fix that after we merged #796

### DIFF
--- a/src/Widgets/TrackRow.vala
+++ b/src/Widgets/TrackRow.vala
@@ -82,6 +82,7 @@ public class Music.TrackRow : Gtk.ListBoxRow {
         row_action_group.add_action (action_remove);
 
         insert_action_group ("trackrow", row_action_group);
+        add_binding_action (Gdk.Key.Delete, Gdk.ModifierType.NO_MODIFIER_MASK, "trackrow.remove", null);
 
         var menu = new Menu ();
         menu.append (_("Remove"), "trackrow.remove");
@@ -106,7 +107,6 @@ public class Music.TrackRow : Gtk.ListBoxRow {
         });
 
         add_controller (right_click);
-
     }
 
     private void update_playing (bool playing) {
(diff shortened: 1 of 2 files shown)


src/Widgets/TrackRow.vala
```diff
@@ -73,6 +73,40 @@ public class Music.TrackRow : Gtk.ListBoxRow {
...
+        menu.append (_("Remove"), "trackrow.remove");
+
+        var context_menu = new Gtk.PopoverMenu.from_model (menu) {
+            halign = Gtk.Align.START,
+            has_arrow = false,
+            position = Gtk.PositionType.BOTTOM
+        };
+        context_menu.set_parent (this);
+
+        var right_click = new Gtk.GestureClick () {
```
danirabbit, line 97: Can we also add long-press here for touch while we're at it :)

danirabbit (changes requested):
Looks good, just one thing

src/Widgets/TrackRow.vala
Oowoosh0: long press should work now

danirabbit (approved):

PR #801, opened 2025-03-25
Introduce a SearchBar

### LINKED ISSUE #794: Search function only shows first occurrence of search term.
### What Happened?

Entering something in the search bar takes you to the first song containing the search term in its title. There is now way to see if there is any other songs in the queue that match the search.

### Steps to Reproduce

1. Latest version from git required. Search is not yet available in the last release.
2. Open multiple files in Music.
3. Searching for something only takes you to the first song containing the search term.

### Expected Behavior

The search should show all songs matching the search term.
Either display the the amount of matches and provide a way to jump from one match to the next or update the list to only contain songs that match the search whenever something is entered into the searchbar.

### OS Version

8.x (Circe)

### Session Type

Secure Session (Wayland)

### Software Version

Compiled from git

### Log Output

```shell

```

### Hardware Info

_No response_

leolost2605:
Fixes #794 

Introduce a search bar widget since we are probably gonna need search in more places if we're going to implement the library.

### DIFF
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -253,6 +253,12 @@ public class Music.MainWindow : Gtk.ApplicationWindow {
                 playback_manager.current_audio = selected_audio;
             }
         });
+
+        search_bar.search_entry.search_changed.connect (() => {
+            if (selection_model.n_items <= 0 && search_bar.search_entry.text != "") {
+                search_placeholder.title = _("No Results for “%s”").printf (search_bar.search_entry.text);
+            }
+        });
     }
 
     public void start_search () {
@@ -339,7 +345,6 @@ public class Music.MainWindow : Gtk.ApplicationWindow {
         }
 
         if (search_bar.search_entry.text != "") {
-            search_placeholder.title = _("No Results for “%s”").printf (search_bar.search_entry.text);
             queue_stack.visible_child = search_placeholder;
         } else {
             queue_stack.visible_child = queue_placeholder;


Oowoosh0 (commented):
Overall it looks good to me. Just a couple of smaller issues I found:
- When nothing matches the search term then the "Queue is Empty" Placeholder gets shown. Probably it would be better to display a placeholder here that reflects that the search didn't return any results.
- There are some performance issues if the queue is larger (couple hundred songs). The UI freezes for a bit on every character entered into the search bar. I guess it is unavoidable that filtering a longer list takes a bit, but ideally the UI should stay responsive.
- If I clear the queue while a search is applied and then add new songs that search is automatically applied again. Maybe it's better to reset the search filter when clearing the queue. But I'm not really sure which of the two would be the expected behaviour for most people.

src/Widgets/SearchBar.vala
```diff
@@ -0,0 +1,53 @@
...
+
+        filter = new Gtk.StringFilter (expression) {
+            ignore_case = true,
+            match_mode = SUBSTRING
+        };
+
+        filter_model = new Gtk.FilterListModel (list_model, filter);
+
+        search_entry = new Gtk.SearchEntry () {
+            placeholder_text = _("Search titles in playlist")
```
ryonakano, line 37: Add this file to POTFILES since it has translatable string.

src/MainWindow.vala
```diff
@@ -55,7 +55,12 @@ public class Music.MainWindow : Gtk.ApplicationWindow {
             icon = new ThemedIcon ("playlist-queue")
         };
 
-        selection_model = new Gtk.SingleSelection (playback_manager.queue_liststore);
+        search_placeholder = new Granite.Placeholder ("") {
+            description = _("Try changing search terms"),
+            icon = new ThemedIcon ("edit-find-symbolic")
+        };
+
+        selection_model = new Gtk.SingleSelection (list_model);
```
ryonakano, line 63: Is there any reason we don't make `search_bar.filter_model` a property and access to it directly here?

I have rarely seen where we use the `out` keyword in the constructor.

danirabbit:
@ryonakano fixed!

ryonakano (approved):

PR #802, opened 2025-03-26
Load cover art more efficiently

Oowoosh0:
Currently Music loads the cover art for every single song added to the queue into memory, no matter if the same cover art was already used by other songs. When opening multiple albums at once this can quickly lead to a lot of memory being wasted.
So I implemented a simple cover cache that makes sure that each individual cover only gets loaded into memory once and is then reused by songs from the same album.

This is still WIP. Just putting it here to avoid that someone else unnecessarily starts working on the same thing.

### DIFF
--- a/src/CoverCache.vala
+++ b/src/CoverCache.vala
@@ -1,0 +1,23 @@
+/*
+ * SPDX-License-Identifier: LGPL-3.0-or-later
+ * SPDX-FileCopyrightText: 2025 elementary, Inc. (https://elementary.io)
+ */
+
+[SingleInstance]
+public class Music.CoverCache : Object {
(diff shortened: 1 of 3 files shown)


teamcons:
If i wanted to continue on this, should i fork your cover-cache branch?
It is out of date but i cant update it since not mine
what is the approach for those cases?

zeebok:
> If i wanted to continue on this, should i fork your cover-cache branch? It is out of date but i cant update it since not mine what is the approach for those cases?

@teamcons I don't know if it is considered the "best" practice but I would fork their branch, add your fixes, and do a PR to merge your change into their branch. If they does not respond after a bit then I guess make a new PR here and we can say it replaces this one.

Issue #792, opened 2025-03-16
Restore last queue


Oowoosh0:
### Problem

I often listen to the same music while working so having to reopen the same couple of folders in the app every day seem tedious.

### Proposal

It would be nice to have an easy way to reload the queue from the last time the app was used. A "Restore last Queue" button could be added underneath the "Queue is Empty" text when the app is opened.

danirabbit:
It's pretty easy to clear the queue so it should probably just be the default behavior. Saving/Loading a list of URIs in gsettings should be pretty straightforward

PR #808, opened 2025-05-27
Attempt at save/restoring queue

teamcons:
Fixes https://github.com/elementary/music/issues/792

This is not ready yet.
The new key in gschema does not seem to install at all. I am trying to figure this out.

-After constructing the window and adding all the signals, when everything is ready, start loading previous queue 
-Handle a close request for the window, the handler slorps everything in the queue in gsettings

alternatively i could change the plumbing to have the app save in realtime whenever the liststore gets changed. This would keep state despite a crash, but it feels gross constantly writing to gsettings

### DIFF
--- a/data/music.gschema.xml
+++ b/data/music.gschema.xml
@@ -12,6 +12,11 @@
       <summary>An index representing the repeat mode</summary>
       <description>An index representing the repeat mode</description>
     </key>
+    <key type="as" name="previous-queue">
+      <default>['']</default>
+      <summary>The queue from last session to restore</summary>
+      <description>An array of strings representing the files played last</description>
+    </key>
 
     <key name="window-height" type="i">
       <default>475</default>
--- a/data/music.gschema.xml
+++ b/data/music.gschema.xml
@@ -13,7 +13,7 @@
       <description>An index representing the repeat mode</description>
     </key>
     <key type="as" name="previous-queue">
-      <default>[]</default>
+      <default>['']</default>
       <summary>The queue from last session to restore</summary>
       <description>An array of strings representing the files played last</description>
     </key>
--- a/data/music.gschema.xml
+++ b/data/music.gschema.xml
@@ -17,6 +17,11 @@
       <summary>The queue from last session to restore</summary>
       <description>An array of strings representing the files played last</description>
     </key>
+    <key type="i" name="index-last-played">
+      <default>0</default>
+      <summary>The item that was playing when the user closed Music</summary>
+      <description>An Integer representing the position in the ListStore of the last played music file</description>
+    </key>
 
     <key name="window-height" type="i">
       <default>475</default>
--- a/data/music.gschema.xml
+++ b/data/music.gschema.xml
@@ -17,10 +17,10 @@
       <summary>The queue from last session to restore</summary>
       <description>An array of strings representing the files played last</description>
     </key>
-    <key type="i" name="index-last-played">
-      <default>0</default>
+    <key type="s" name="uri-last-played">
+      <default>''</default>
       <summary>The item that was playing when the user closed Music</summary>
-      <description>An Integer representing the position in the ListStore of the last played music file</description>
+      <description>A string representing the uri of the last played music file</description>
     </key>
 
     <key name="window-height" type="i">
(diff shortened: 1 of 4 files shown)


data/music.gschema.xml
```diff
@@ -12,6 +12,11 @@
       <summary>An index representing the repeat mode</summary>
       <description>An index representing the repeat mode</description>
     </key>
+    <key type="as" name="previous-queue">
+      <default>[]</default>
```
danirabbit, line 16: I think this isn't valid. It probably should be 
```suggestion
      <default>['']</default>
```

src/MainWindow.vala
```diff
@@ -243,6 +243,21 @@ public class Music.MainWindow : Gtk.ApplicationWindow {
...
+        var last_session_uri = settings.get_strv ("previous-queue");
+        var last_session_files = new File[last_session_uri.length];
+
+        foreach (var uri in last_session_uri) {
+            var file = File.new_for_uri (uri);
+            last_session_files += file;
+        }
+
+        var files_to_play = Application.loop_through_files (last_session_files);
+        PlaybackManager.get_default ().queue_files (files_to_play);
```
danirabbit, line 260: I think this should probably be done during `startup` in the application or when the playbackmanager is constructed. It doesn't really have anything to do with the MainWindow UI, so probably it shouldn't happen here

src/MainWindow.vala
```diff
@@ -243,6 +243,21 @@ public class Music.MainWindow : Gtk.ApplicationWindow {
                 playback_manager.current_audio = selected_audio;
             }
         });
+
+        close_request.connect (e => {
+            return before_destroy ();
```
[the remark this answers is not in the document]
teamcons: Fair. Working on it

teamcons:
Okay
I need to figure how to start playing a track in the liststore at a specific position (Int)
I will do that tomorrow. But saving/restoring works

data/music.gschema.xml
teamcons: I was very unsure which one between both. Because [''] is technically an array with an empty array
While [] is an empty array
Thank you

src/MainWindow.vala
teamcons: I first tried at the end of playback construct - But it ended with the Gtk.Window displaying an empty queue while playing music
So i moved this in the startup, after window is presented and everything is ready to play.

src/Application.vala
```diff
@@ -114,6 +114,17 @@ public class Music.Application : Gtk.Application {
...
+
+        // Restoring the queue overwrites the last played. So we need to retrieve it before
+        var? uri_last_played = settings.get_string ("uri-last-played");
+
+        // This needs to be done after window is constructed
+        // Else music plays but the queue seems empty
+        PlaybackManager.get_default ().restore_queue ();
+
+        settings.set_string ("uri-last-played", uri_last_played);
+        PlaybackManager.get_default ().restore_last_played ();
```
danirabbit, line 126: Is there a reason to have this be two separate functions? We always want to restore both the previous queue and the previously played audio right?

teamcons:
So i can save the stupid last played track. Reliably.
but i cannot figure out how to tell playback to say "please play this AudioObject ive set as current_audio"

src/PlaybackManager.vala
```diff
@@ -477,6 +469,28 @@ public class Music.PlaybackManager : Object {
...
+                progress_timer = 0;
+            }
+        }
+
+        update_next_previous_sensitivity ();
+
+        var play_pause_action = (SimpleAction) GLib.Application.get_default ().lookup_action (Application.ACTION_PLAY_PAUSE);
+        play_pause_action.set_enabled (current_audio != null);
+
+        settings.set_string ("uri-last-played", current_audio.uri);
```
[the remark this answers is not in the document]
teamcons: "uri-last-played" gets overwritten before being used. But apart from that, yes i suppose, but do not forget this is a draft - i want first to have something that works (im failing at it), before making it clean.

teamcons:
Yes. I cant figure this out.
Giving up on this PR. If anyone want to pick up from here please do.

src/Application.vala
danirabbit: Oh yeah that makes sense. Thanks for clarifying!

teamcons:
Should i close the PR ? Sorry im useless i thought i could tackle this, but it just doesnt care whatever i set as current audio or call, it will always reset to 0 after the activate()

PR #809, opened 2025-05-30
Move lambdas into handlers

teamcons:
This adds code clarity and makes sure we have dedicated handlers for whatever comes next.


danirabbit (approved):
Looks good. Nice little refactoring. Very clean :)

PR #810, opened 2025-05-30
Introduce save and restore methods

teamcons:
The initial draft had no actual own save handler.
having a method for one but not the other felt awkward

### DIFF
--- a/src/PlaybackManager.vala
+++ b/src/PlaybackManager.vala
@@ -38,6 +38,7 @@ public class Music.PlaybackManager : Object {
     private PlaybackManager () {}
 
     construct {
+        settings = new Settings ("io.elementary.music");
         queue_liststore = new ListStore (typeof (AudioObject));
 
         playbin = Gst.ElementFactory.make ("playbin", "playbin");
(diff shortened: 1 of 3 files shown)


src/PlaybackManager.vala
```diff
@@ -455,4 +456,30 @@ public class Music.PlaybackManager : Object {
         var play_pause_action = (SimpleAction) GLib.Application.get_default ().lookup_action (Application.ACTION_PLAY_PAUSE);
         play_pause_action.set_enabled (current_audio != null);
     }
+
+        public void save_queue () {
```
danirabbit, line 460: Doesn't look like this is being called in this branch.  In #808 this appears to be part of `on_items_changed ()` which makes sense to me

danirabbit (changes requested):

src/PlaybackManager.vala
```diff
@@ -455,4 +457,30 @@ public class Music.PlaybackManager : Object {
...
```
danirabbit, line 461: Seems like there's some funny indentation here? Also since this isn't called from outside of this class, it should be private

danirabbit (changes requested):

danirabbit (approved):
Looks good and works as expected. Nice job!

teamcons:
Ya! Aint we cool? I think id wanna try at some point too!

PR #812, opened 2025-05-31
Refresh screenshot to reflect current state

teamcons:
-Showcase the search bar being used
-Clear queue button is present
-Show app state with artist and cover art
-UI in english for the occasion

Petty change but that saves someone else a fifteen minutes.


danirabbit (changes requested):
Looks like this screenshot is missing shadows

danirabbit:
Hmm it looks like the screenshot app is messing up for some reason. Keyboard shortcut works correctly though!

danirabbit (approved):
Looks good, thank you!

PR #814, opened 2025-06-09
Restore last played

teamcons:
I lost the PR ? Leaving this as draft, still dont know
Final item to fix https://github.com/elementary/music/issues/792

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -115,9 +115,18 @@ public class Music.Application : Gtk.Application {
 
         settings.bind ("window-maximized", main_window, "maximized", SettingsBindFlags.SET);
 
+
+        // Restoring the queue overwrites the last played. So we need to retrieve it before
+        var? uri_last_played = settings.get_string ("uri-last-played");
+
         // This needs to be done after window is constructed
         // Else music plays but the queue seems empty
         playback_manager.restore_queue ();
+
+        settings.set_string ("uri-last-played", uri_last_played);
+        playback_manager.restore_last_played ();
+        print(uri_last_played);
+
     }
 
     private static File[] list_directory (string directory) {
--- a/data/music.gschema.xml
+++ b/data/music.gschema.xml
@@ -17,6 +17,11 @@
       <summary>The queue from last session to restore</summary>
       <description>An array of strings representing the files played last</description>
     </key>
+    <key type="s" name="uri-last-played">
+      <default>''</default>
+      <summary>The item that was playing when the user closed Music</summary>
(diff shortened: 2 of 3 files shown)


src/Views/NowPlayingView.vala
```diff
@@ -108,10 +108,16 @@ public class Music.NowPlayingView : Gtk.Box {
 
         playback_manager.notify["current-audio"].connect (() => {
             if (playback_manager.current_audio != null) {
-                playback_manager.current_audio.bind_property ("artist", artist_label, "label", BindingFlags.SYNC_CREATE);
-                playback_manager.current_audio.bind_property ("title", title_label, "label", BindingFlags.SYNC_CREATE);
-                playback_manager.current_audio.bind_property ("duration", seekbar, "playback-duration", BindingFlags.SYNC_CREATE);
-                playback_manager.current_audio.bind_property ("texture", album_image.image, "paintable", BindingFlags.SYNC_CREATE);
+                artist_label.label = playback_manager.current_audio.artist;
+                title_label.label = playback_manager.current_audio.title;
+                album_image.image.paintable = playback_manager.current_audio.texture;
```
danirabbit, line 113: Is there a reason you changed these from a bind?

The reason these were bound previously is we want to make sure to update if metadata changes after the audio is set, such as if it's discovered late or changes.

danirabbit (changes requested):

src/Views/NowPlayingView.vala
teamcons: Yea, so i quickly explained, but basically i found that NowPlayingView was being weird/fucky. It would sometimes show the restored track info correctly, sometimes not
I tried to play a bit around, but only found this solution where it would follow for sure whatever was playing.

teamcons: I put the binds back. The issue reproduced, then everything functions normally

Could you please test, and see if you get issues with metadata please?

image

src/Application.vala
```diff
@@ -115,8 +115,6 @@ public class Music.Application : Gtk.Application {
 
         settings.bind ("window-maximized", main_window, "maximized", SettingsBindFlags.SET);
 
-        // This needs to be done after window is constructed
-        // Else music plays but the queue seems empty
```
danirabbit, line 119: I think this was a good comment actually. It's good to provide the reason we don't do this in `playback_manager`'s construct block

danirabbit (approved):
This looks great and works as expected. Nice job! 🚀

PR #823, opened 2025-08-13
Metainfo: invert dark/light brand colors

danirabbit:
Somehow this was done backwards:

## BEFORE

Screenshot from 2025-08-13 09 56 18
Screenshot from 2025-08-13 09 56 34


wpkelso (approved):
Looks better to me. A part of me wonders if this was originally done for contrast purposes, but I think it looks a bit silly if it was

PR #824, opened 2025-08-13
PlaybackManager: create actions internally

danirabbit:
Setup playbackmanager actions inside the playbackmanager. Makes sure playback manager functions are private so they're only used through actions. Manage state internally. Simplify setup


ryonakano:
@danirabbit Hmm it's difficult to judge, but maybe they would be better to be kept in Application?

These constants themselves are no longer used in Application but the actions invoked by using these constants are still registered to it. So thinking about where we use them, I feel the former (current) code makes sense and hide internal implementation that the handler of the actions are actually no longer in Application but in PlaybackManager:

```vala
        var play_button = new Gtk.Button () {
            action_name = Application.ACTION_PREFIX + Application.ACTION_PLAY_PAUSE,
            child = play_pause_image,
            halign = Gtk.Align.CENTER,
        };
```

rather than:

```vala
        var play_button = new Gtk.Button () {
            // I think this implicitly make the users of this action, us programmers, to be aware of the internal implementation
            action_name = Application.ACTION_PREFIX + PlaybackManager.ACTION_PLAY_PAUSE,
            child = play_pause_image,
            halign = Gtk.Align.CENTER,
        };
```

src/PlaybackManager.vala
```diff
@@ -60,6 +65,35 @@ public class Music.PlaybackManager : Object {
...
+        play_pause_action.set_enabled (false);
+        previous_action.set_enabled (false);
+        shuffle_action.set_enabled (false);
+
+        var action_group = GLib.Application.get_default ();
+        action_group.add_action (clear_action);
+        action_group.add_action (next_action);
+        action_group.add_action (play_pause_action);
+        action_group.add_action (previous_action);
+        action_group.add_action (shuffle_action);
```
ryonakano, line 94: I feel this is strange with the variable name because the interface that has `add_action()` is not `GLib.ActionGroup` but `GLib.ActionMap`.

But anyways I think we don't need to think about which interface has it and simply name as `app`. Also explicit `unowned`

```suggestion
        unowned var app = GLib.Application.get_default ();
        app.add_action (clear_action);
        app.add_action (next_action);
        app.add_action (play_pause_action);
        app.add_action (previous_action);
        app.add_action (shuffle_action);
```

src/PlaybackManager.vala
```diff
@@ -409,10 +442,7 @@ public class Music.PlaybackManager : Object {
 
         var default_application = GLib.Application.get_default ();
```
ryonakano, line 443: This seems to be no longer used.

```
../src/PlaybackManager.vala:443.13-443.31: warning: Local variable `default_application' declared but never used
  443 |         var default_application = GLib.Application.get_default ();
      |             ^~~~~~~~~~~~~~~~~~~                                   
```

ryonakano (commented):
Confirmed these actions still works. One small nitpicking otherwise LGTM.

ryonakano (approved):
Thanks for the cleanup!  :broom: :sparkles:

PR #825, opened 2025-08-13
Release 8.1.0

danirabbit:
I'd like to get in:
- [x] #823 
- [x] #814 
- [x] #801


teamcons:
Do i have time to revise the m3u support? should be able to get around it this week. Arrived back home.

danirabbit:
@teamcons id like to get this release out as soon as possible so we can release a new stable ISO with the latest HWE. We can always make another release of music in a few weeks or at the end of the month, but I don't want to block folks who need the newer HWE on app features

Issue #832, opened 2025-10-31
Open folders directly from the Music app


o-alquimista:
### Problem

Currently, the only way to select and recursively add all music files contained in a folder is through the file manager, using the "Open In" context menu. The app window offers no equivalent functionality, as it only lets you select one or more files, not folders.

### Proposal

On [Amberol](https://apps.gnome.org/Amberol/) — which is a very similar player — you can press the _Add Folder_ button to select any folder and include all playable audio files contained within into the playlist.

<img width="768" height="665" alt="Image" src="https://github.com/user-attachments/assets/ad04f3e0-52c4-46b3-9039-03a723823439" />

This would require recursively traversing the directory structure to gather all music files inside. Note that I organize my music in an `Artist > Album` structure, and I often want to play all albums from an artist. Amberol lets me select the Artist folder to include everything inside.

teamcons:
@o-alquimista should Music add only the files directly in the folder, or should it loop through subfolders till everythings there

flodavid:
Actually you could already select a folder (using CTRL + Left Clic, because the File Dialog does not respect "Double Click to Navigate" setting of Files) to recursively import a directory. But I created a PR to add a button that eases the process.
The first element is no longer selected when you use this new button, and validating with "Open" recursively imports the current directory.

dakshshrivastav:
example of how this could be done:-

pictured `byte - music player`; indexes folder and then implements playlist/queue logic instead of directly doing just that.

<img width="532" height="786" alt="Image" src="https://github.com/user-attachments/assets/6d856b88-83b7-48fb-8f9b-29254ead85e6" />

<img width="532" height="786" alt="Image" src="https://github.com/user-attachments/assets/de31dc3a-0214-4da6-b70e-eb25d1e73893" />

<img width="532" height="786" alt="Image" src="https://github.com/user-attachments/assets/8123cb8d-cb7b-462d-a6b5-b6e5c9d413ae" />

---

here's the folder structure it got this out of (byte scans recursively)

<details>
<summary>ls output of folder structure in `~/Music`</summary>

`~/Music > Folder > Artist > Album > .mp3`

```
❯ ls
 whole-lotta-leaks  'YouTube Music'
❯ ls YouTube\ Music
'$NOT'                               Future                                      Nedaj
'¥$, Playboi Carti, & Don Toliver'  'Gajendra Verma'                             Nemzzz
'1400 Vault'                         Gakuen                                     'Nephew Obama'
[...]

PR #838, opened 2026-03-24
Adds an "Open Folder…" button to open folders only

flodavid:
- Eases the import of directories by never selecting an element. This way, the "Open" button opens the current directory instead of import the selected file(s)/folder(s).
  - Validating with "Open" imports the current directory and its subdirectories (if no element has been manually selected)
- Prevents selecting individual files
- Assign CTRL + SHIFT + O shortcut to the Open Folder dialog

Addresses #832

### DIFF
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -131,6 +131,9 @@ public class Music.Application : Gtk.Application {
         return elements;
     }
 
+    /**
+     * Flattens files and those contained in (sub)directories
+     */
     public static File[] loop_through_files (File[] files) {
         // All of these will be returned later in bulk
         File[] elements = {};
--- a/src/MainWindow.vala
+++ b/src/MainWindow.vala
@@ -6,6 +6,7 @@
 public class Music.MainWindow : Gtk.ApplicationWindow {
     public const string ACTION_PREFIX = "win.";
     public const string ACTION_OPEN = "action-open";
+    public const string ACTION_OPEN_FOLDER = "action-open-folder";
 
     private QueueView queue_view;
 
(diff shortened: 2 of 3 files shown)


jeremypw (commented):
The code looks good and works as expected.  However I would like input from the UX team as to whether the addition of an extra button is the correct solution.  It is already possible to select folders in the existing "Open File" dialog provided the "All Files" filter is selected.  It may be better to have the default filter select music files *and* folders and then an additional button is not required.

zeebok:
> The code looks good and works as expected. However I would like input from the UX team as to whether the addition of an extra button is the correct solution. It is already possible to select folders in the existing "Open File" dialog provided the "All Files" filter is selected. It may be better to have the default filter select music files _and_ folders and then an additional button is not required.

The issue though is that it is frustrating to select more than one folder. Clicking a folder opens it. Plus picking more than one from the `Open Files...` dialog doesn't work.

jeremypw:
> Plus picking more than one from the Open Files... dialog doesn't work.
Yes, selecting multiple folders needs fixing in any case.

zeebok:
I think if there is going to be a new button, maybe having a kebab menu (the three vertical dots) button next to `Open Files...` and within that can be the `Open Folders...` button. It could additionally hold the Save Queue to Playlist button that I would probably add to finish up playlist support.

Maybe @elementary/ux has some strong opinion here?

flodavid:
One issue for me with the "Open files..." button is that the dialog does not respect the "Files" app "Double-click to Navigate" setting. So I can not select an element without holding the "Control" key.
The problem is more with the dialog than the Music app, but it would be more involved to fix the core issue.

flodavid:
> Plus picking more than one from the `Open Files...` dialog doesn't work.

I do not get what is not working. I can open select multiple files/folders by holding control and are all imported.

jeremypw:
> One issue for me with the "Open files..." button is that the dialog does not respect the "Files" app "Double-click to Navigate" setting.

See https://github.com/elementary/files/pull/2711  The intention is to eventually completely replace the current Gtk filechooser with one similar to Files.

jeremypw:
> I do not get what is not working. I can open select multiple files/folders by holding control and are all imported.

I think it does not work in `main` but it does work in this PR.

Issue #408, opened 2018-08-16
Media playing position is not restored properly


jeremypw:
If the app is closed with a medium partly played (and continue playing turned off) then when reopened the medium position is returned to zero i.e. it starts playing from the beginning.

danrabbit:
Thanks for your report! We're doing a from-scratch rewrite of Music and this issue no longer applies to the new code base. I'm closing this report since the issue can't be reproduced in the new version of Music.

PR #839, opened 2026-03-29
Save and restore playback progress

flodavid:
Addresses #408, even if this was closed because of the app rewrite, I think it is still relevant.

- Add progress in seconds to the gschema to save it every 2 seconds


src/PlaybackManager.vala
flodavid, line 62: I just removed this line as it duplicates line 57

teamcons:
i appreciate the intent, but wouldnt this be a bit aggressive in read/write and SSD wear, for no benefit?

half of the time, half of the song will have already played for a new session. So people will get annoyed and restart it anyway.
We listen to songs _a lot_ and it plays on loop for background atmosphere or music when GMing TTRPGs.

flodavid:
That’s why I only saved progress every two seconds. This makes disk writes pretty low. And this feature is common on other players of many platforms.
I do not get what is the point of saving the current song if the progress is not saved. You risk playing the same track at its beginning on multiple launch ; I find this annoying.

PR #840, opened 2026-03-29
Add support for playlists

### LINKED ISSUE #319: How to import playlist?
Previously, I created a playlist and exported it for future use. But now, after I changed distro, I can't find how to import my exported playlist.

zeebok:
Closes #319 

Builds off of the previous PR but addresses some outstanding changes and moves the playlist logic to an Object. Currently only supports reading .m3u extension playlists.


src/Application.vala
```diff
@@ -148,6 +150,16 @@ public class Music.Application : Gtk.Application {
...
+            else if (PlaylistObject.is_playlist (file)) {
+                PlaylistObject playlist = new PlaylistObject.with_file (file);
+
+                foreach (File playlist_file in playlist.get_uri_list ()) {
+                    elements += playlist_file;
+                }
+
+                // Don't add the playlist file itself
+                continue;
+            }
```
zeebok, line 163: Thinking about the code more, I am curious if this should maybe live in `PlaybackManager.queue_file` and check if the playlist file is a supported type, that way anything that falls under "audio/x-mpegurl" but is not an m3u causes the invalid file toast instead of being added to the queue as a broken AudioObject?

teamcons: [audio/x-mpegurl should be m3u or m3u8](https://mime-type.com/audio/x-mpegurl/)

dont confuse with [audio/mpeg](https://mime-type.com/audio/mpeg/)

unless mime-type.com speaks lies, but mimetype.io seems to say the same thing

teamcons:
> Closes #319
> 
> Builds off of the previous PR but addresses some outstanding changes and moves the playlist logic to an Object. Currently only supports reading .m3u extension playlists.

thank you for overtaking. I wasnt feeling that one anymore.

src/Application.vala
zeebok: Well for some reason playlist extensions of other programs (like a fake .pls file) will be visible with that filter? Maybe I shouldn't worry about it until it is actually a problem that happens with other people lol

src/PlaylistObject.vala
```diff
@@ -0,0 +1,74 @@
...
+
+        return mimetype == "audio/x-mpegurl";
+    }
+
+    public File[]? get_uri_list () {
+        return uri_list;
+    }
+
+    private void load_playlist (File playlist) {
+        FileInputStream fis = playlist.read ();
```
jeremypw: This causes an uncaught error warning if `playlist` does not yet exist - which happens when creating a new playlist for example.

jeremypw (changes requested):
Basically works, but I noticed one quirk: You can load the same playlist multiple times resulting multiple copies in the queue. Shouldn't loading a playlist overwrite the existing one?  Also if you close and reopen the app with such a duplicated playlist then the duplicate entries appear as full file paths not the name of the track.

I'll do some more testing.

src/PlaylistObject.vala
zeebok: How do you load a playlist when creating it?

jeremypw: I did not do anything special - just used the save as playlist action. 
This results in the terminal warning
```
** (io.elementary.music:28593): CRITICAL **: 18:50:24.991: file ../src/PlaylistObject.vala: line 47: uncaught error: Error opening file /home/jeremy/TestFolder/TestMusic/New Playlist.m3u: No such file or directory (g-io-error-quark, 1)

```

zeebok: Oh I think I see the cause. Will fix this.

zeebok:
> Basically works, but I noticed one quirk: You can load the same playlist multiple times resulting multiple copies in the queue. Shouldn't loading a playlist overwrite the existing one? Also if you close and reopen the app with such a duplicated playlist then the duplicate entries appear as full file paths not the name of the track.
> 
> I'll do some more testing.

In regards to this, would your PR for checking for duplicates would resolve this?

jeremypw:
> In regards to this, would your PR for checking for duplicates would resolve this?

Yes, it should do. It was due to pre-existing issue, not this PR.

jeremypw (changes requested):
This works pretty well now. The only thing I noticed is that when saving a playlist, existing playlists did not appear when the Playlist filter was active, only when the "All Files" filter was active.  This is because you used `add_pattern` with a mime-type not a glob pattern. So it should be `add_pattern ("*.m3u")` for example, and/or `playlist_filter.add_mime_type ("audio/mpegurl");`

jeremypw (approved):
Looks good.  Let's go!

Next thing will  be to expose this action in the UI e.g. with headerbar button.

PR #841, opened 2026-04-07
Respect ReplayGain meta data

flodavid:
Applies the rgvolume plugin as filter to support track gain/peak, album gain/peak

The plugin is part of GStreamer's `gst-plugins-good`. See more: https://gstreamer.freedesktop.org/documentation/replaygain/rgvolume.html


zeebok (approved):

zeebok:
I don't think I have any music files that have this metadata to test, but based on the link you provided it seems like this is all you'd need for this to work!

Issue #255, opened 2018-01-12
Reorder the queue


elegaanz:
Make it possible to drag items of the queue to another place (or at least, add a "Play after the current song" context menu option).

Oowoosh0:
This is being worked on in the context of #779 and will be added soon.

PR #844, opened 2026-05-14
Sort queue on track number if available

### LINKED ISSUE #767: Chronogical file order when importing a folder
### Problem

When a folder is imported, files are scattered randomly in the playlist (2, 5, 1) instead of being sorted by mp3 track number (1, 2, 3)

### Proposal

Fixing it is as simple as importing it in order of track number in mp3 metadata.

### Prior Art (Optional)

_No response_

jeremypw:
Fixes #767


zeebok:
Out of curiosity, if you were to open multiple albums at once via Files how does this handle the ordering? Currently if I multi-select album folders and use Open In Music, the songs are not in order within the album, but they are grouped by album. Would be cool if it grouped by album and in album order!

jeremypw:
@zeebok I agree the ordering of the queue needs to be more sophisticated/customizable (see #255,  #767, #779).  This is just a start.

